PR #12, opened 2019-07-11
update gomod to use 1.0.0 of signalfx-go module

austburn:


cory-signalfx:
Hey @austburn!

Forgive me, but I'm working this out in #15, which should land in the next few business days.

PR #24, opened 2019-07-17
Rework some default handling

cory-signalfx:
# Summary

Simplify the handling of some of the `MaxXX` defaults we're using to fix up the lack of proper "default + zero" handling in Terraform

# Motivation

The pointer changes in #22 had me experimenting if we needed this. We still do, but can clean it up a bit. This fixes up some problems I was having with new charts getting junk max/min values too.

### DIFF
--- a/signalfx/resource_signalfx_single_value_chart.go
+++ b/signalfx/resource_signalfx_single_value_chart.go
@@ -243,7 +243,7 @@ func singlevaluechartCreate(d *schema.ResourceData, meta interface{}) error {
 	payload := getPayloadSingleValueChart(d)
 
 	debugOutput, _ := json.Marshal(payload)
-	log.Printf("[DEBUG] SignalFx: Create List Chart Payload: %s", string(debugOutput))
+	log.Printf("[DEBUG] SignalFx: Create Single Value Chart Payload: %s", string(debugOutput))
 
 	chart, err := config.Client.CreateChart(payload)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


justinskariah:
`high_watermark` and `low_watermark` are being set to `340282346638528860000000000000000000000` and `-340282346638528860000000000000000000000` respectively by default. Which causes all `signalfx_time_chart` resource to be updated because of `high_watermark` and `low_watermark` changes.

```
      - axis_left {
          - high_watermark = 179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 -> null
          - label          = "%" -> null
          - low_watermark  = 179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 -> null
          - max_value      = 110 -> null
          - min_value      = 0 -> null
        }
      + axis_left {
          + high_watermark = 340282346638528860000000000000000000000
          + label          = "%"
          + low_watermark  = -340282346638528860000000000000000000000
          + max_value      = 110
          + min_value      = 0
        }
```

Please let me know if you need further information on this. Should I raise a bug for this?

cory-signalfx:
Hey @justinskariah! These values are an unfortunate necessity because Terraform can't differentiate between an unset value and 0… and since 0 is a valid place to put a high or low watermark, we have to choose hilariously large defaults.

What's bad is if this is affecting a resource after an apply. It's possible that this PR caused you have have an unclean plan after an upgrade of the provider, but running apply should make it go away.

In other words, please open an issue if — after running `apply` — you see this diff in the plan persisting.

justinskariah:
Thanks @cory-signalfx, I have raised #35

PR #36, opened 2019-07-25
Updating docs for Opsgenie notifications. 

austburn:
From what I see in the code credentialName is not supported. `Team` is the only supported `responderType` as per the API docs. I linked the docs for detector creation as well


cory-signalfx:
Thanks @austburn!

Issue #28, opened 2019-07-18
Bad notification strings should not panic


cory-signalfx:
From #25 we see it's possible to exceed the index of the array due to having an invalid string. We should validate these to avoid future panics.

cory-signalfx:
Fixed in #48

PR #46, opened 2019-08-06
Add notification validation

cory-signalfx:
# Summary

Adds validation for notification strings.

# Motivation

The notification strings used in detectors get the job done, but they are a minefield for users. It's easier to leave things out or have errors in the fields. As such, we need to validate to help the user get things right. This also prevents crashed from index out of bounds problems on the string split.


nickchappell:
👍 for this. I run into issues with this all the time.

cory-signalfx:
That was a fast comment! 🐎

cory-signalfx:
Fixes #28

Issue #66, opened 2019-08-28
Floating point numbers sometimes come back with precision errors


dmason-atlassian:
Sometimes the state fetched from remote will have a floating point rounding error, causing the comparison to the local copy to be mismatched, even though it's not, and thus triggers a change. I've noticed this especially on low_watermark configuration for an axis.

mattthebig:
I'm also getting this problem, I never have a clean plan after applying, horrible for the developer workflow. @cory-signalfx Any progress on getting that PR merged?

cory-signalfx:
Hey @mattthebig, I'm sorry to hear this is interfering with your work. I've not merged this PR mostly because I can't reproduce it and I'm leery of making such a change that isn't reproducible.

I've rebased the PR, could you or @dmason-atlassian perchance try this branch out with the places you are seeing this and report back?

cory-signalfx:
This should be fixed in #174!

PR #69, opened 2019-08-30
Loosen our comparison tolerances for floats

cory-signalfx:
# Summary

Use `MaxInt32` for range defaults and value comparisons.

# Motivation

In #66 it was reported that floating point is being floating point. Due to floating point fuckery sometimes we get an unclean plan. I considered switching to `string` but that doesn't fix the `signalfx-go` library which is going to still return 0 defaults.* To that end, this loosens the tolerance of our checks and just ropes off anything above `MaxInt32` or below `-MaxInt32`. This seems safe because `float64` has > 32 bits of mantissa so it can accurately represent this number with no fear of serialization problems.

Thanks, Go!

* We can't make 0 unusable for these values because of color ranges, which heavily use 0.


cory-signalfx:
Tagging in @keitwb for some Go sanity checks. Any better ideas?

signalfx/util.go
```diff
@@ -208,7 +208,7 @@ func getColorScaleOptions(d *schema.ResourceData) []*chart.SecondaryVisualizatio
 }
 
 func getValueUsingMaxFloatAsDefault(v float64) *float32 {
-	if v >= math.MaxFloat32 || v <= -math.MaxFloat32 {
```
keitwb, line 182: The function name seems misleading, it doesn't "use max float as default" anywhere in this function.

keitwb, line 182: Also is there any good reason not to just clamp the value to max/min float32 if it is OOB?

cory-signalfx: Changed!

keitwb (approved):
LGTM.  Hacky but sane given the limitations of Terraform config.

cory-signalfx:
This should be fixed in #174

Issue #72, opened 2019-09-09
"one step" AWS account integration


kimor79:
Currently, to add an AWS integration, one needs to use the UI to create the integration. This returns an external ID (and SignalFx account ID) which is then used in the trust policy of the IAM role and a subsequent call to the integration API. There doesn't exist a way to do this all within terraform. And if the current signalfx_aws_integration were to be able to do both calls to the API, there would be a circular dependency in terraform:
```
resource "aws_iam_role" "main" {
  ...
  assume_role_policy = <<EOF
// ${signalfx_aws_integration.main.external_id} and ${signalfx_aws_integration.main.signalfx_account_id} (which doesn't seem to be exported) are needed here
EOF
}

resource "signalfx_aws_integration" "main" {
  auth_method = "ExternalId"
  ...
  role_arn    = "${aws_iam_role.main.arn}"
}
```

We would like to complete the entire integration within terraform, something like:
```
resource "signalfx_aws_integration_pre" "main" {
  auth_method = "ExternalId"  
  name = "${var.account_id}"
  ...
}

resource "aws_iam_role" "main" {
  ...
  assume_role_policy = <<EOF
  // ${signalfx_aws_integration_pre.main.external_id}
  // ${signalfx_aws_integration_pre.main.account_id}
EOF
}

resource "signalfx_aws_integration" "main" {
  integration_id = "${signalfx_aws_integration_pre.main.id}"
  ...
  role_arn = "${aws_iam_role.main.arn}"
}
```

cory-signalfx:
Thanks for making this, @kimor79! This is a great idea, and I appreciate you writing it up for me. I don't think I'll get to this today, but plan to get on it next week.

cory-signalfx:
Alrighty, having taken a close look at this I think this is how it can work:

* A new resource `signalfx_aws_integration_pre` which is created and stores the integration's ID. It can be deleted, but has no meaningful update operation. It also has no properties, other than the computed external id.
* Add a `signalfx_aws_integration_external` which is basically the same as the current, but it *requires* an external id, which can be referenced from the `signalfx_aws_integration_pre`. This resource's create call will actually be an update. It will update normally, but have no delete. It will require the id and external_id from `signalfx_aws_integration_pre`.
* Add a `signalfx_aws_integration_token` which handles the other auth mechanism.
* Deprecate `signalfx_aws_integration`.

Does that sound sane?

cory-signalfx:
[...]

PR #81, opened 2019-09-17
Multi-step AWS resources for non-circular integration

cory-signalfx:
# Summary

* Adds a `signalfx_aws_external_integration` resource
* Adds a `signalfx_aws_token_integration` resource
* Adjusts the `signalfx_aws_integration` to utilize the above resources
* Updates the documentation reflect all of the above

# Motivation

Currently there is a circular dependency when managing AWS resources. Because the user must acquire an `external_id` via a `POST` call but cannot yet provide an ARN, we need two resources. The first resource will provide us with an `external_id` we can reference in the resource. This allows a setup like:

```
resource "signalfx_aws_external_integration" "foo" {
  name = "AWS Integration"
}

resource "aws_iam_role" "role_for_signalfx" {
  // …
}

resource "signalfx_aws_integration" "bar" {
  integration_id = ""${signalfx_aws_external_integration.id}"
  external_id = "${signalfx_aws_external_integration.external_id}"
  role_arn = "${aws_iam_role.role_for_signalfx.arn}"
}
```

# Notes

See #72


signalfx/resource_signalfx_aws_external_integration.go
```diff
@@ -0,0 +1,119 @@
...
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: "Name of the integration",
+			},
+			"external_id": &schema.Schema{
+				Type:        schema.TypeString,
+				Computed:    true,
+				Sensitive:   true,
+				Description: "The SignalFx-generated AWS account ID to use with an AWS integration.",
```
kimor79, line 26: isn't the SignalFx _AWS account_ ID different than the _external_ ID?

cory-signalfx: Yup!

Issue #97, opened 2019-10-03
Can't use more than 2 color_scales


nickchappell:
* Affected version: 4.8.3

I tried adding some color scales to a single value chart resource with three different levels, green at or above 99.95, yellow from 99.5 thru 99.95 and red below 99.5. TF code below:

```bash
  color_scale {
    gte   = 99.95
    color = "green"
  }

  color_scale {
    gte   = 99.5
    color = "yellow"
  }

  color_scale {
    lt    = 99.5
    color = "magenta"
  }
```

I get the following error:

```bash
Error: Unexpected status code: 400: {
  "code" : 400,
  "message" : "Color scales missing range between 99.500000 and 99.950000"
}
```

Trying different combos of gt, gte and lt/lte give the same result.

Side note: Keeping the `color_scale`s above but with `color_by` set to something other than "Scale" doesn't result in any errors at all. Maybe some extra validation could be added to point out that `color_by` needs to be set if `color_scale` options are present?

doctornkz:
You should use the structure like that:
```
color_by                = "Scale"
color_scale = [
{ color = "green", gt = 3, lte = 6 }, 
{ color = "yellow", gt = 1, lte = 3 }, 
{ color = "red", gte = 0, lte = 1 }
]
```
and yes, you have non-strong situation after 99.95 - because first and the second statements are working together. you should pass `lt=99.95` into second one.

cory-signalfx:
Hey @nickchappell!

@doctornkz suggestion looks correct to me. Let us know if that helps?

Your suggestion about `color_by` is great. I will leave this open to add that too!

nickchappell:
@cory-signalfx @doctornkz's suggestion doesn't work. The range errors are gone, but now, I get:

```bash
Error: Unsupported argument

  on sfx.tf line 51, in resource "signalfx_list_chart" "success_rate_30d":
  51:   color_scale = [

An argument named "color_scale" is not expected here. Did you mean to define a
block of type "color_scale"?
```
If I go back to the multiple `color_scale {}` blocks like I had, but with the extra `gte/lte` range specs, it does work.

Working code:

```bash
  color_scale {
    color = "green"
    gt    = 95
    lte   = 100
  }

  color_scale {
    color = "yellow"
    gt    = 90
    lte   = 95
  }

  color_scale {
    color = "magenta"
    gte   = 0
    lte   = 90
  }
```

doctornkz:
[...]

PR #99, opened 2019-10-07
List charts complain if color scales are specified without the correct color_by

cory-signalfx:
# Summary

If `color_scale` is used with a `color_by` other than `Scale`, complain.

# Motivation

Raised as an idea in #97.

### DIFF
--- a/signalfx/resource_signalfx_list_chart.go
+++ b/signalfx/resource_signalfx_list_chart.go
@@ -2,6 +2,7 @@ package signalfx
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
 	"math"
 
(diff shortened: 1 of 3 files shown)


cory-signalfx:
Tagging in @keitwb

signalfx/resource_signalfx_list_chart.go
```diff
@@ -251,6 +255,10 @@ func getListChartOptions(d *schema.ResourceData) *chart.Options {
 			if colorScaleOptions := getColorScaleOptions(d); len(colorScaleOptions) > 0 {
 				options.ColorScale2 = colorScaleOptions
 			}
+		} else {
+			if val, ok := d.GetOk("color_scale"); ok && val != nil {
+				return nil, fmt.Errorf("Using `color_scale` without `color_by = \"Range\"` has no effect")
```
keitwb: This error condition is confusing.  This `else` block is hit when `color_by != "Scale"` so potentially couldn't `color_by == "Range"` here?

cory-signalfx: Oops!

keitwb (approved):

PR #100, opened 2019-10-14
Fix column width in grid schema

ImDevinC:
Fixes the column width setting. Documentation says that 1-12 is supported, but code only supports 1-11. Also added a unit test for this.


cory-signalfx:
Thanks @ImDevinC!

Issue #101, opened 2019-10-15
`grid` on dashboard resources doesn't allow multiple grids


tur-nr:
We are trying to upgrade the provider to version 4 but all of our grids have broken 😢 

```terraform
resource "signalfx_dashboard" "detections" {
  name            = "Dashboard"
  dashboard_group = signalfx_dashboard_group.group.id

  // 2x1 grid (count=2)
  grid {
    chart_ids = ["1", "2"]
    width  = 6
  }
  
  // 3x3 grid (count=9)
  grid {
    chart_ids = ["3", "4", "5", "6", "7", "8", "9", "10", "11"]
    width  = 4
  }
}
```

What is the reason behind a single grid? As before this worked in version 3.

Related to #80

cory-signalfx:
Hello @tur-nr. There's a reason, but it's not very interesting and doesn't do anything to fix your problems.

As I said in #80, I think I can fix this up. Thanks for the test case. I'll set it up locally and see what I can do.

cory-signalfx:
Oops, this was closed in #102!

PR #102, opened 2019-10-15
Adjust behavior of layouts to allow more than one instance.

### LINKED ISSUE #80: `column` on dashboard resources doesn't allow mulitple columns
I've been porting over my SFX dashboard TF code fro the Yelp provider to this official version, but am getting the following error:

```bash
Error: too many items for attribute "column"; cannot have more than 1
```

The code that causes this looks like roughly like this and worked with the Yelp version of the provider (resources names are anonymized and the `chart_ids` are converted to TF v0.12-compatible syntax) :

```bash
  column {
    chart_ids = "${signalfx_time_chart.1day_slo.*.id}"
    column    = 0
    width     = 4
    height    = 2
  }

  column {
    chart_ids = "${signalfx_time_chart.1week_slo.*.id}"
    column    = 4
    width     = 4
    height    = 2
  }

  column {
    chart_ids = "${signalfx_time_chart.1month_slo.*.id}"
    column    = 8
    width     = 4
    height    = 2

  }
```

The dashboard resource code has `MaxItems` set to 1: https://github.com/terraform-providers/terraform-provider-signalfx/blob/master/signalfx/resource_signalfx_dashboard.go#L131

The old Yelp version of the provider, however, doesn't have a `MaxItems` set, and allows mulitple columns: https://github.com/Yelp/terraform-provider-signalform/blob/master/src/terraform-provider-signalform/signalform/dashboard.go#L162

I'm assuming the same issue would happen with multiple `grid`s, though I don't have any dashboards laid out like that.

cory-signalfx:
# Summary

Allow more than one `grid` and `column`.

# Motivation

Fixes #80 and #101

# Notes

This violates normal Terraform behavior, as it doesn't really examine the charts upon receiving the API response. This is a tradeoff to allow much more convenient layout. This is a necessity to retain this feature.

The exclusivity of each type (`layout`, `grid`, and `chart`) is retained, as the API has no way to sort them out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ BUG FIXES:
 
 * resource/dashboard: Corrected validation of chart widths, allowing 12. Thanks [@ImDevinC](https://github.com/ImDevinC) [#100](https://github.com/terraform-providers/terraform-provider-signalfx/pull/100)
 
+IMPROVEMENTS:
+
+* resource/dashboard: Multiple instances of `column` and `grid` can now be used. [#102](https://github.com/terraform-providers/terraform-provider-signalfx/pull/102)
+
 ## 4.9.0 (October 10, 2019)
 
 FEATURES:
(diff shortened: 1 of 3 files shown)


keitwb (approved):
LGTM.  Is this worthy of being in the CHANGELOG though?

keitwb:
> LGTM. Is this worthy of being in the CHANGELOG though?

Nevermind, you were too quick.

Issue #108, opened 2019-11-05
Add resource for muting rules


cory-signalfx:
There's currently no way to manage muting rules, we should add this!

geoffhigginbottom:
My use case for this is that when a new set of infra is deployed into an environment it can cause an alert storm as things come on line and settle.  The ask is to be able to temporarily mute detectors for the new infra for a defined period of time, to allow it to settle before then re-enabling them.

PR #110, opened 2019-11-06
Alert Muting Rule Resource

cory-signalfx:
# Summary

Adds a resource for alert muting rules

# Motivation

#108


signalfx/resource_signalfx_alert_muting_rule.go
```diff
@@ -0,0 +1,256 @@
...
+	alertmuting "github.com/signalfx/signalfx-go/alertmuting"
+)
+
+func alertMutingRuleResource() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"description": &schema.Schema{
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Description of the rule.",
```
keitwb: This is somewhat pedantic but none of these descriptions are sentences -- so the period really should just be omitted.

signalfx/resource_signalfx_alert_muting_rule.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+		if err := d.Set("detectors", detectors); err != nil {
+			return err
+		}
+		// The API changes `startTime` to be >= the current
+		// timestamp at the time of the API call. This means
+		// it will pretty much never agree with what the user specified.
+		// To accomodate this we will store the "effective" start time
+		// as a computed attribute, then…
+		if err := d.Set("effective_start_time", amr.StartTime); err != nil {
```
keitwb, line 206: You divided `amr.StopTime` by 1000 (I'm guessing converting milliseconds to seconds), but don't here. Below the effective start time seems to be swappable with start_time in seconds. Does this need a `/ 1000` here too?

signalfx/resource_signalfx_alert_muting_rule.go
cory-signalfx: Fixed!

signalfx/resource_signalfx_alert_muting_rule.go
cory-signalfx: Nope. I left a comment about that in the Schema, but the gist is that this field is not user-facing and therefore it's not necessary to do any seconds conversion on it. We only do the seconds conversion on other fields because humans typically thing of Unix time as seconds. No need to do it on this one. (Hopefully the comment helps that for future readers?)

keitwb (approved):

Issue #116, opened 2019-11-26
Panic (and subsequent terraform crash) due to nil pointer dereference


DethAriel:
Hi folks! I've recently run into this issue during `terraform apply` with existing state. The version of terraform I'm running is `0.12.13`, and the `terraform-provider-signalfx` is of version `4.9.2`.

I'd be happy to provide additional information, but was hoping that the trace might be sufficient already :)

```
panic: runtime error: invalid memory address or nil pointer dereference
plugin.terraform-provider-signalfx_v4.9.2_x4: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xdf95b3]
plugin.terraform-provider-signalfx_v4.9.2_x4: 
plugin.terraform-provider-signalfx_v4.9.2_x4: goroutine 1065 [running]:
plugin.terraform-provider-signalfx_v4.9.2_x4: github.com/terraform-providers/terraform-provider-signalfx/signalfx.dashboardAPIToTF(0xc00012a0e0, 0xc0001b4500, 0x3, 0xedc6c0)
plugin.terraform-provider-signalfx_v4.9.2_x4: 	/opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-signalfx/signalfx/resource_signalfx_dashboard.go:877 +0x433
plugin.terraform-provider-signalfx_v4.9.2_x4: 	/opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-signalfx/signalfx/resource_signalfx_dashboard.go:714 +0x1cd
plugin.terraform-provider-signalfx_v4.9.2_x4: github.com/terraform-providers/terraform-provider-signalfx/signalfx.dashboardRead(0xc00012a0e0, 0xea46c0, 0xc0004d8500, 0xc00012a0e0, 0x0)
plugin.terraform-provider-signalfx_v4.9.2_x4: github.com/terraform-providers/terraform-provider-signalfx/vendor/github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).RefreshWithoutUpgrade(0xc000125180, 0xc000517720, 0xea46c0, 0xc0004d8500, 0xc0000a7dc8, 0x0, 0x0)
[...]

PR #120, opened 2019-12-04
Fix crash for some dashboard reads

cory-signalfx:
# Summary

Fixes a crash for dashboards with no `EventSignal` field.

# Motivation

In #116 the crash was reported. Fixes that!


cory-signalfx:
Fixes #116

Issue #128, opened 2019-12-17
Can't use credentialId in Webhook notifications for Detector rules


shwin:
I'd like to use a webhook notification in a signalFX detector, but only provide the `credentialId`.

The official API doc  (https://developers.signalfx.com/detectors_reference.html#operation/Create%20Single%20Detector, expand rules and notifications and click on webhook) says this is possible:
![image](https://user-images.githubusercontent.com/1700973/70954374-cb987f80-2021-11ea-8b91-46090aaa7371.png)

So in Terraform, I have a detector with a rule like:
```
rule {
    detect_label = "critical"
    severity     = "Critical"

    notifications = [
      "Webhook,SOMEID",
    ]
  }
```

That gives me an error:
`Error: Invalid Webhook notification string, please consult the documentation (not enough parts)`

Ok, fine, I don't mind providing the URL:
```
  notifications = [
      "Webhook,SOMEID,,http://mywebsite.com",
    ]
```
But I get: `"message" : "Either url or credentialId must be defined, both were defined."`

If I do it like
```
  notifications = [
      "Webhook,SOMEID,,",
    ]
```

I get `Error: Invalid Webhook URL ""`

So, is there any way I can use `credentialId` in terraform?

shwin:
I figure this is just a bug in this logic: https://github.com/terraform-providers/terraform-provider-signalfx/blob/master/signalfx/notifications.go#L205

```
case WebhookNotificationType:
		if len(parts) != 4 {
			errs = append(errs, fmt.Errorf("Invalid Webhook notification string, please consult the documentation (not enough parts)"))
			return
		}
		_, err := url.ParseRequestURI(parts[3])
		if err != nil {
			errs = append(errs, fmt.Errorf("Invalid Webhook URL %q", parts[2]))
			return
		}
```

We should probably not require 4 parts if they provide `credentialId`, since the rest of the parts are ignored if you do.

cory-signalfx:
Hey @shwin! Could you look over #129 and pass along any feedback?

cory-signalfx:
Closing this as I plan to ship a new provider today with #129 in it!

cory-signalfx:
Should go out today!

PR #129, opened 2019-12-17
Fix notification webhook validation

cory-signalfx:
# Summary
Improves Webhook validation for notifications

# Motivation
In #128  it was reported that Webhooks can be specified as either a credential ID — which uses an existing Webhook "integration" — or defined inline with a URL and secret. Neither of these scenarios worked because the validation required 4 "parts" to the input.

This PR changes this behavior to check for both scenarios. It also attempts to complain helpfully.

# Notes

This should really be replaced with a `notification` resource that specifies and validates these silly fields rather than this goofball comma thing that is a pain for users.

### DIFF
--- a/signalfx/notifications.go
+++ b/signalfx/notifications.go
@@ -69,8 +69,6 @@ func getNotifyStringFromAPI(not *notification.Notification) (string, error) {
 	default:
 		return "", fmt.Errorf("Unknown notification type %q", nt)
 	}
-
-	return "", nil
 }
 
 func getNotifications(tfNotifications []interface{}) ([]*notification.Notification, error) {
@@ -170,7 +168,8 @@ func getNotifications(tfNotifications []interface{}) ([]*notification.Notificati
 
 func validateNotification(val interface{}, key string) (warns []string, errs []error) {
 	parts := strings.Split(val.(string), ",")
-	if len(parts) < 2 {
+	partCount := len(parts)
+	if partCount < 2 {
 		errs = append(errs, fmt.Errorf("Invalid notification string, not enough commas"))
 		return
 	}
@@ -184,12 +183,12 @@ func validateNotification(val interface{}, key string) (warns []string, errs []e
 			return
 		}
 	case OpsgenieNotificationType:
-		if len(parts) != 5 {
+		if partCount != 5 {
 			errs = append(errs, fmt.Errorf("Invalid OpsGenie notification string, please consult the documentation (not enough parts)"))
 			return
 		}
 	case SlackNotificationType:
-		if len(parts) != 3 {
+		if partCount != 3 {
 			errs = append(errs, fmt.Errorf("Invalid Slack notification string, please consult the documentation (not enough parts)"))
 			return
 		}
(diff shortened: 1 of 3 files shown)


signalfx/notifications.go
```diff
@@ -198,19 +197,36 @@ func validateNotification(val interface{}, key string) (warns []string, errs []e
...
 		}
 	case VictorOpsNotificationType:
-		if len(parts) != 3 {
+		if partCount != 3 {
 			errs = append(errs, fmt.Errorf("Invalid VictorOps notification string, please consult the documentation (not enough parts)"))
 			return
 		}
 	case WebhookNotificationType:
-		if len(parts) != 4 {
+		if partCount != 4 {
```
shwin, line 205: should we require 4 parts at all if they provide the credentialId?

like having to do `Webhook,<credentialId>,,` seems silly versus just `Webhook,<credentialId>`.

Though! I understand if we just always want 4 parts and figure things out from there.

Ie, I'm just suggesting moving the `parts[1] != "" ` check "higher", and then check the rest based on that, but I'm ok if that makes things more complicated.

shwin: one other point of unsolicited feedback :p

this pattern of "4 things in a string but it matters the order" is not very terraform-y; I'd love if some day we had proper structs like:
```
notification {
   type = "Webhook"
   Url = "mywebsite.com"
}
```

etc.

website/docs/r/detector.html.markdown
```diff
@@ -117,8 +117,16 @@ notifications = ["TeamEmail,teamId"]
 
 ### Webhook
 
+~> **NOTE** You need to include all the commas even if you only use a credential id below.
```
shwin, line 120: 🙌 for doc updates!

shwin (approved):
this looks great! 

left some comments more for my edification than anything else :)

shwin:
💯 to 
>goofball comma thing  

lol

shwin:
@cory-signalfx sorry that I couldn't find this otherwise but; what's the overall release process for this repo?

it looks like when this is merged, teamcity or something publishes a new provider somewhere? and then my existing terraform should pull down that version?

i use the provider like
```
provider "signalfx" {
}
```

so I figure it always just uses the latest, but would love to know how a PR actually gets out there :)

signalfx/notifications.go
cory-signalfx: Yeah, I didn't want to get into a "variable number of commas" situation. Perhaps I'm just being conservative, but I'd rather invest in an improved `notification` resource than such a change. (But, to be fair, I started the change by going that direction.)

cory-signalfx:
Good question, @shwin. I feel this is a bit vague in Terraform so I don't mind the question. :)

Since you've not provided a constraint in that `provider` block — which, to be clear, you can do if you want — then you can run `terraform init --upgrade` to upgrade. It will **not** upgrade without action from you. It wants such a change to be intentional.

I'll be pinging HashiCorp and they'll do the release for me when they get to it. Usually same day! Look for it tomorrow morning maybe?

cory-signalfx:
Update, this will probably go out in a new Terraform build tomorrow, not today. It took me a bit more time to get #125 reviewed internally for spiritual accuracy. :)

PR #144, opened 2020-01-29
Add config options to prefix dashboard groups and detectors

cory-signalfx:
# Summary

Adds configuration options that allow prefixing Dashboard Groups and Detectors with a fixed string.

# Motivation

It's common for users to prefix assets created by Terraform with some form of string to signal it's… Terraformedness so that users know. This can also be done with tags or permissions, but neither are obvious in SignalFx's UI. This allows a prefix, like:

```
provider "signalfx" {
  auth_token = "${var.signalfx_auth_token}"
  dashboard_group_name_prefix = "[TF]"
  detector_name_prefix = "[TF]"
}
```


cory-signalfx:
This doesn't work. Terraform only looks at the config for state, so it makes unclean plans trying to revert the prefix, or refuses to add a prefix if one is configured after asset creation.

Issue #151, opened 2020-02-21
Error when creating couple data_links through terraform for a same dashboard_id and same context_dashboard_id with the different property_name ( which is not expected behavior )


saikrishna547:
Why can't we create 2 data_links with the different property_name and having same context_dashboard_id and dashboard_id(target dashboard) saying "Cannot create multiple cross links for the same property value combination" ( which is not the expected behavior )

Code 1)

```
resource "signalfx_data_link" "mydatalink_overview_1" {
  context_dashboard_id = "${signalfx_dashboard.overview.id}"
  property_name = "AWSUniqueId"
  target_signalfx_dashboard {
    is_default = true
    name = "Overview awsuniqueid to host dashboard"
    dashboard_group_id = "${signalfx_dashboard_group.xyz_dashboard_group.id}"
    dashboard_id = "${signalfx_dashboard.host.id}"
  }
}
```
Code 2)
```
resource "signalfx_data_link" "mydatalink_overview_6" {
  context_dashboard_id = "${signalfx_dashboard.overview.id}"
  property_name = "host"
  target_signalfx_dashboard {
    is_default = true
    name = "Overview containerID data link to Docker Container dashboard"
    dashboard_group_id = "${signalfx_dashboard_group.xyz_dashboard_group.id}"
    dashboard_id = "${signalfx_dashboard.host.id}"
  }
}
```

These things can be created in UI manually, but not in terraform ( which is not right for anyone, we can have to different properties leading to a same dashboard from the same context_dashboard).

cory-signalfx:
Hey @saikrishna547, this doesn't look like a problem within Terraform to me, as I'm able to create data links using your provided examples (i.e. two links with the same targets, but different property names). I also double checked the API code that verifies and agree that it does work!

I suspect you've either got another definition elsewhere in your Terraform or someone has created a data link in the UI that is not governed by Terraform. Keep in mind that the API does not differentiate between assets created with Terraform and those made by hand!

cory-signalfx:
This is (perhaps confusingly) a dupe of #150 

I believe this was a bug wherein the property name wasn't be set!

PR #154, opened 2020-02-25
Fix property name only for data links

### LINKED ISSUE #150: Couple issues with the terraform data_links( couldn't get the property name in the UI ) after successful plan and apply.
Could get the property name in the UI for the below code or for any of the code, It could get the property value even showing us in the terraform plan and also apply.

resource "signalfx_data_link" "mydatalink_overview_1" {
  context_dashboard_id = "${signalfx_dashboard.overview.id}"
  property_name = "AWSUniqueId"
  target_signalfx_dashboard {
    is_default = true
    name = "Overview awsuniqueid to host dashboard"
    dashboard_group_id = "${signalfx_dashboard_group.xyz_dashboard_group.id}"
    dashboard_id = "${signalfx_dashboard.host.id}"
  }
}

Screen Shot 2020-02-21 at 9 47 27 AM
Screen Shot 2020-02-21 at 9 47 42 AM
Screen Shot 2020-02-21 at 9 48 14 AM

ghost:
property_name was not being set for datalinks if property_value was missing.

Also property_name is required for dashboard targets.

Fixes #150, #151

### DIFF
--- a/signalfx/resource_signalfx_data_link.go
+++ b/signalfx/resource_signalfx_data_link.go
@@ -161,11 +161,19 @@ func getPayloadDataLink(d *schema.ResourceData) (*datalink.CreateUpdateDataLinkR
 		}
 	}
 
+	if name, ok := d.GetOk("property_name"); ok {
+		dataLink.PropertyName = name.(string)
+	}
+
 	if val, ok := d.GetOk("context_dashboard_id"); ok {
 		dataLink.ContextId = val.(string)
 	}
 
 	if val, ok := d.GetOk("target_signalfx_dashboard"); ok {
+		if name, ok := d.GetOk("property_name"); !ok || name.(string) == "" {
+			return nil, fmt.Errorf("Must supply a property_name when using target_signalfx_dashboard")
+		}
+
 		sfxDashes := val.(*schema.Set).List()
 
 		for _, tfLink := range sfxDashes {
--- a/signalfx/resource_signalfx_data_link.go
+++ b/signalfx/resource_signalfx_data_link.go
@@ -83,7 +83,7 @@ func dataLinkResource() *schema.Resource {
 							Optional:    true,
 							Description: "Designates the format of minimumTimeWindow in the same data link target object.",
 							ValidateFunc: validation.StringInSlice([]string{
-								string(datalink.ISO8601), string(datalink.Epoch),
+								string(datalink.ISO8601), string(datalink.Epoch), string(datalink.EpochSeconds),
 							}, false),
 						},
 						"url": &schema.Schema{
--- a/signalfx/resource_signalfx_data_link.go
+++ b/signalfx/resource_signalfx_data_link.go
@@ -152,17 +152,15 @@ func dataLinkResource() *schema.Resource {
 func getPayloadDataLink(d *schema.ResourceData) (*datalink.CreateUpdateDataLinkRequest, error) {
 	dataLink := &datalink.CreateUpdateDataLinkRequest{}
 
-	if val, ok := d.GetOk("property_value"); ok {
-		if name, ok := d.GetOk("property_name"); ok {
-			dataLink.PropertyValue = val.(string)
-			dataLink.PropertyName = name.(string)
-		} else {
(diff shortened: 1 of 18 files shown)


signalfx/resource_signalfx_data_link.go
```diff
@@ -161,11 +161,19 @@ func getPayloadDataLink(d *schema.ResourceData) (*datalink.CreateUpdateDataLinkR
...
+	if name, ok := d.GetOk("property_name"); ok {
+		dataLink.PropertyName = name.(string)
+	}
+
 	if val, ok := d.GetOk("context_dashboard_id"); ok {
 		dataLink.ContextId = val.(string)
 	}
 
 	if val, ok := d.GetOk("target_signalfx_dashboard"); ok {
+		if name, ok := d.GetOk("property_name"); !ok || name.(string) == "" {
```
ghost: Do we need to strip name before doing a whitespace comparison? Or is checking for == "" not even needed?

signalfx/resource_signalfx_data_link.go
```diff
@@ -161,11 +161,19 @@ func getPayloadDataLink(d *schema.ResourceData) (*datalink.CreateUpdateDataLinkR
 		}
 	}
 
+	if name, ok := d.GetOk("property_name"); ok {
```
cory-signalfx: I feel this can be simplified by putting this above the check for property value. Something like:

```
if name, ok := d.GetOk("property_name"); ok {
	dataLink.PropertyName = name.(string)
}
if val, ok := d.GetOk("property_value"); ok {
	if dataLink.PropertyName == "" {
		return dataLink, fmt.Errorf("Must supply a property_name when supplying a property_value")
        }
	dataLink.PropertyValue = val.(string)
}
```

This way we're not setting `propertyName` in two places and the logic easily allows only a name, or a value and a name.

signalfx/resource_signalfx_data_link.go
cory-signalfx: Wait, is it a thing that property name is required for links to dashboards, but not the other types?!

signalfx/resource_signalfx_data_link.go
cory-signalfx: Also, the [docs](https://developers.signalfx.com/datalinks_reference.html#tag/Create-Data-Link) say `propertyName` is required? Should we change ours from `Optional` to `Required`?

cory-signalfx (changes requested):

cory-signalfx:
This is also the problem behind #151, fwict.

signalfx/resource_signalfx_data_link.go
ghost: Digging into the code I see that GetOk already checks if it's a zero value. So I guess the == "" isn't needed unless extra whitespace is a concern. But maybe TF handles this higher up in the stack so that whitespace only gets normalized to empty string?

ghost: Yep. I discovered this in the UI. If you click through the different data link types you'll see the trigger options are different for the different types. Only dashboard targets require a property name best I can tell.

cory-signalfx: Fascinating!

Ok, will look over this with that in mind.

signalfx/resource_signalfx_data_link.go
cory-signalfx: Strike the required thing above, I see it's Complicated™.

signalfx/resource_signalfx_data_link.go
cory-signalfx: Ok, I would just look at `dataLink.PropertyName` in the above code, since you've already verified it there. Also, we should probably make a DOCS ticket that asks to clarify this since I don't think this is explained in the API docs.

cory-signalfx (changes requested):

signalfx/resource_signalfx_data_link.go
ghost: https://signalfuse.atlassian.net/browse/DOCS-1143

cory-signalfx (approved):
Yay! Merge away!

PR #155, opened 2020-02-25
Fix property name not being set on data links

cory-signalfx:
# Summary

Property value on Data links wasn't being set.

# Motivation

Flagged in #150 and related to #151


cory-signalfx:
@jaysfx is working on this in #154

Issue #165, opened 2020-03-04
Some missing pieces or errors in detectors documentation


xp-1000:
Hello,

Reading the documentation from https://www.terraform.io/docs/providers/signalfx/r/detector.html I believe I have detected some errors (i.e. missing count) but I could be wrong, you will tell me in the PR.

xp-1000:
Here is the PR: https://github.com/terraform-providers/terraform-provider-signalfx/pull/166

cory-signalfx:
Thanks, merged!

PR #166, opened 2020-03-04
#165 bunch of improvements on detector documentation

xp-1000:
please, let me now if I did some errors of interpretation I will fix them.


cory-signalfx:
Excellent, thanks!

(I had intended to remove all of the `count` stuff in that example, but failed. So putting it back is probably best.)

PR #170, opened 2020-03-11
Fix namespace sync rules with missing filters

cory-signalfx:
# Summary

Correctly process `namespace_sync_rules` with no filters.

# Motivation

When using an `aws_integration` with a mix of `namespace_sync_rules` with and *without* filters, the provider was ignoring those *without* the filter. It did this under the mistaken impression that the handling of `services` would pick those up. It doesn't! The `services` and `namespace_sync_rules` options are mutually exclusive. This meant that any sync rule with only a namespace, such as:

```
namespace_sync_rule {
  namespace = "AWS/EBS"
}
```

would be ignored on the response and cause an unclean plan. 😿


xp-1000:
Thanks :tada:

PR #183, opened 2020-03-26
Respect standard environment variables for HTTP proxies

### LINKED ISSUE #182: Allow configuring an HTTPS proxy
There is currently no way to do this. My terraform modules must run in an environment where outbound HTTPS requests go through a proxy, so this prevents me from using terraform-provider-signalfx.

RafeKettler:
This should fix #182; ProxyFromEnvironment will use a proxy
if any of the standard HTTP/HTTPS_PROXY env vars are set, and
will use no proxy if these env vars are not set, or if NO_PROXY
is set.


cory-signalfx:
Fascinating, thanks so much for this!

PR #187, opened 2020-04-05
Update list_chart.html.markdown

ebutleratlassian:
`plot_label` appears unused by default list charts, however `sf_metric` in terraform corresponds to "Plot Name" in signalfx and `sf_originatingMetric` corresponds to "metric (sf_metric)" in the signalfx UI which matches the first argument to `data` in signalflow, an argument called "metric". confusing!
Does this correlation change based on some configuration? This is what I saw while working with my own list chart.

sf_originatingMetric -> metric (sf_metric) -> signal label, first argument to data in signalflow "metric"
sf_metric -> Plot Name -> plot label or display name

### DIFF
--- a/website/docs/r/list_chart.html.markdown
+++ b/website/docs/r/list_chart.html.markdown
@@ -77,7 +77,7 @@ The following arguments are supported in the resource block:
     * `value_prefix`, `value_suffix` - (Optional) Arbitrary prefix/suffix to display with the value of this plot.
 * `legend_fields_to_hide` - (Optional) List of properties that should not be displayed in the chart legend (i.e. dimension names). All the properties are visible by default. Deprecated, please use `legend_options_fields`.
 * `legend_options_fields` - (Optional) List of property names and enabled flags that should be displayed in the data table for the chart, in the order provided. This option cannot be used with `legend_fields_to_hide`.
-    * `property` The name of the property to display. Note the special values of `plot_label` (corresponding with the API's `sf_metric`) which shows the label of the time series `publish()` and `metric` (corresponding with the API's `sf_originatingMetric`) that shows the name of the metric for the time series being displayed.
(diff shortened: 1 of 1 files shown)


cory-signalfx:
The intention here is that you would use one of the these three values:
* The name of a dimension or property, etc
* The value `metric` which is translated to `sf_originatingMetric` for the API and yields the name of the metric
* The value `plot_label` which is translated  to `sf_metric` for the API and yields the name of the `publish(…)` block and which corresponds to "Plot Name" in the UI

A complication is that if you use `sf_originatingMetric` or `sf_metric` as values it will skip the translation. It feels like maybe that's what you're seeing?

As for list charts, they use the function `getLegendOptions` which covers the logic above.

Can you help me understand what is confusing, I'd definitely like to clear that up!

ebutleratlassian:
Hey @cory-signalfx, thanks for replying. The issue I saw was that in my usage I noticed that 
```
resource "signalfx_list_chart" "my_list" {
    // other config...
    legend_options_fields {
        property = "plot_label"
        enabled  = false
    }
}
```
had no effect, but when I changed `property` to `sf_metric` directly it did.
Is that normal?

BTW I'm using terraform@0.12.24 and signalfx provider@4.18.6

cory-signalfx:
Ah, ok! Lemme double check that on my end. First glance says it should work…

cory-signalfx:
Ok @ebutleratlassian, sorry for the confusion. I had it in my head that the two different field options worked the same and they do not! Your doc patch is correct!

ebutleratlassian:
Awesome, thanks for confirming it for me, I'm new to terraform and was hoping I hadn't gotten myself confused 😄 👍

cory-signalfx:
No worries, thank you doubly for investing the time to learn and contribute. It's really helpful!

PR #218, opened 2020-06-04
Document named_token on GCP integration

pdecat:
This PR is about documenting the new `named_token` attribute for all integrations supporting it.

Right now, it only updates the GCP integration to discuss if it's the right way to do it, hence the draft PR.

Will generalize once okay.

### DIFF
--- a/website/docs/r/gcp_integration.html.markdown
+++ b/website/docs/r/gcp_integration.html.markdown
@@ -31,10 +31,49 @@ resource "signalfx_gcp_integration" "gcp_myteam" {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


cory-signalfx:
For now we'd like to leave this undocumented, as there's still some optimization to be done internally. :(

cory-signalfx:
(We can leave this here for later tho.)

pdecat:
Ok, noted.

Should we avoid using it for now then? I'm already conducting tests one of our organizations.

cory-signalfx:
I don't think it's a problem. You should maybe let your SE know just in case?

PR #221, opened 2020-06-26
Adjust user agent

### LINKED ISSUE #208: Dashboard creation too slow since 4.19
I am creating dashboards with over 30 graphs. This used to take about 2s per dashboard. Since v4.19 it takes 37s. I am unable to add more graphs to a dashboard because at some point I get HTTP 504 from the server (which used to be OK before). Example below:

```
resource "signalfx_dashboard_group" "group0" {
  name = "Example Dashboard Group"
}

resource "signalfx_dashboard" "dashboard0" {
  name = "Example Dashboard"
  dashboard_group = signalfx_dashboard_group.group0.id

  column {
    width = 12
    height = 1
    column = 0
    chart_ids = signalfx_text_chart.text0.*.id
  }
}

resource "signalfx_text_chart" "text0" {
  count = 60
  name  = "Text"

  markdown = <<-EOF
    Hello world
  EOF
}
```
Verified with 4.21, same issue.

cory-signalfx:
# Summary
Adjust the user agent to include the standard Go string.

# Motivation
It's bloody magic. For some reason this makes it ridiculously faster and fixes #208

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 4.24.0 (Unreleased)
+## 4.23.2 (Unreleased)
+
+IMPROVEMENTS:
+* provider: Adjusted the HTTP User-Agent supplied by the provider when making API calls. [#221](https://github.com/terraform-providers/terraform-provider-signalfx/pull/221)
+
 ## 4.23.1 (June 10, 2020)
 
 IMPROVEMENTS:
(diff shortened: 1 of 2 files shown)


cory-signalfx:
I'm going to go ahead and merge this because it merely reverts to behavior before `v1.19.0`, wherein we used the default Go User-Agent. This allows users to avoid the slowdown while we sort this out on our end.

Issue #233, opened 2020-08-06
[Question] use_get_metric_data_method useless ?


xp-1000:
Hello,

I would like to use https://registry.terraform.io/providers/splunk-terraform/signalfx/latest/docs/resources/aws_integration#use_get_metric_data_method everytime which is more efficient.

I am asking myself if this option is still useful now because I do not see it in the API doc : https://developers.signalfx.com/integrations_reference.html#operation/Create%20Integration

So I suppose, this is a legacy option when default method still was the old one and may be the default method (and only one available) is the get_metric_data.

Could you confirm this please ?

also it seems poll rate should be a "free" value between 60 and 600 included (and not and choice between both of them).

Also again (sorry to ask multiple questions into one issue ^^) I see the following option in API doc : 

```
enableCheckLargeVolume | boolean (AWS CloudWatch Large Data Volume Flag)  Controls how SignalFx checks for large amounts of data for this AWS integration. If true, SignalFx monitors the amount of data coming in from the integration.
```

I think I understand the purpose and this is interesting but I am not sure what is the factual impact (I suppose there should add new aws metric(s) with some stats about crawler but it is not clear and I don't see any more detailled doc).

Thanks !

cory-signalfx:
Hey @xp-1000! No worries on the multiple questions, I'm happy to help. In order:

1. The `use_get_metric_data_method` does work, but I would encourage you to contact SignalFx support about it's use and implications. It is not documented intentionally, from what I understand.
2. It does look as though `poll_rate` allows arbitrary values between 60 and 600, I'll work on that!
3. The `enableCheckLargeVolume` is not currently supported in Terraform, but I can add it. Expect a pull request!

cory-signalfx:
Hey @xp-1000, [v4.26.0](https://github.com/splunk-terraform/terraform-provider-signalfx/releases/tag/v4.26.0) should be available soon from Terraform and fix items 2 and 3 in your list. I am closing this issue now since 1 requires a support discussion.  Thanks for flagging!

PR #234, opened 2020-08-06
AWS Poll Rate Range and Check Large Volume Support

cory-signalfx:
# Summary
* Allow AWS `poll_rate` to be a range.
* Add AWS large volume flag.
* Remove some unused code.

# Motivation
In #233 is was pointed out that the underlying API for this changed without my knowledge, so updating it to allow a range rather than two fixed values. In addition, nobody had yet asked for large volume support, so adding that!


cory-signalfx:
Relevant to #233

PR #246, opened 2020-09-14
Rename provider module path to match GH organisation

stack72:
This matches what other providers have done:

e.g. https://github.com/digitalocean/terraform-provider-digitalocean/pull/463


jrcamp:
Thanks!

PR #257, opened 2020-10-02
Team links

jrcamp:
FEATURES:
* resource/signalfx_team_links: This new resource can be used to manage linking of detectors and dashboard groups to teams. Multiple links can be used to a single team.

BREAKING CHANGES:
* resource/signalfx_team: The recently added dashboard group and detector entries have been removed from teams and must now be managed with the `signalfx_team_links` resource.

### DIFF
--- a/signalfx/resource_signalfx_team_links.go
+++ b/signalfx/resource_signalfx_team_links.go
@@ -1,0 +1,183 @@
+package signalfx
+
+import (
+	"context"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func teamLinksResource() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"team": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "Team ID to link the configured dashboard groups or detectos to.",
+			},
+			"dashboard_groups": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "Dashboard group IDs to link to the team.",
+				Optional:    true,
+			},
+			"detectors": {
+				Type:        schema.TypeSet,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "Detector IDs to link to the team.",
+				Optional:    true,
+			},
+		},
+		Create: teamLinksCreate,
+		Read:   teamLinksRead,
+		Update: teamLinksUpdate,
+		Delete: teamLinksDelete,
+		Exists: teamLinksExists,
+	}
+}
+
+func teamLinksExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	return true, nil
+}
+
+func teamLinksRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*signalfxConfig)
+	teamId := d.Get("team").(string)
+	res, err := config.Client.GetTeam(context.TODO(), teamId)
+	if err != nil {
+		if strings.Contains(err.Error(), "404") {
+			d.SetId("")
+		}
+		return err
+	}
+
+	_, new := d.GetChange("detectors")
+	detectors := new.(*schema.Set)
+
+	dets := make([]interface{}, len(res.Detectors))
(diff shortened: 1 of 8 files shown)


signalfx/resource_signalfx_team_links.go
```diff
@@ -0,0 +1,183 @@
...
+		Create: teamLinksCreate,
+		Read:   teamLinksRead,
+		Update: teamLinksUpdate,
+		Delete: teamLinksDelete,
+		Exists: teamLinksExists,
+	}
+}
+
+func teamLinksExists(d *schema.ResourceData, meta interface{}) (bool, error) {
+	return true, nil
```
keitwb, line 41: Is this always true because the links are such an abstract concept or what?

signalfx/resource_signalfx_team_links.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+func teamLinksRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*signalfxConfig)
+	teamId := d.Get("team").(string)
+	res, err := config.Client.GetTeam(context.TODO(), teamId)
+	if err != nil {
+		if strings.Contains(err.Error(), "404") {
```
keitwb, line 49: Setting id to blank string only on 404s is not entirely clear to me.  Can you explain the reasoning behind this in a comment?

signalfx/resource_signalfx_team_links.go
```diff
@@ -0,0 +1,183 @@
...
+	for i, detector := range res.Detectors {
+		dets[i] = detector
+	}
+
+	teamDets := schema.NewSet(schema.HashString, dets)
+	// XXX: For some reason the same sets are hashing to different values (maybe to interface{} vs string types).
+	// Rehash the set so intersection, etc. will work.
+	detectors = schema.NewSet(schema.HashString, detectors.List())
+
+	if err := d.Set("detectors", teamDets.Intersection(detectors)); err != nil {
```
keitwb, line 68: The logic leading up to here is not very easy to follow.  Maybe rename some stuff or move some stuff out to helper functions.  It really isn't obvious what two things are being intersected here.

signalfx/resource_signalfx_team_links.go
```diff
@@ -0,0 +1,183 @@
...
+func teamLinksUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*signalfxConfig)
+	team := d.Get("team").(string)
+
+	old, new := d.GetChange("detectors")
+	oldD := old.(*schema.Set)
+	newD := new.(*schema.Set)
+
+	removed := oldD.Difference(newD)
+	added := newD.Difference(oldD)
```
keitwb, line 104: The naming here is much easier to follow than in the read function.

signalfx/resource_signalfx_team_links.go
```diff
@@ -0,0 +1,183 @@
...
+		if err := config.Client.LinkDashboardGroupToTeam(context.TODO(), team, det); err != nil {
+			return err
+		}
+	}
+
+	if err := d.Set("dashboard_groups", new); err != nil {
+		return err
+	}
+
+	d.SetId("0")
```
keitwb, line 157: What does an id of `"0"` mean?

signalfx/resource_signalfx_team_links.go
jrcamp: Right since there's no actual resource on the backend that gets created or deleted this value doesn't really matter. More details here https://www.terraform.io/docs/plugins/provider.html#resources

Actually looks like it's optional so I can just not even implement it since it has no use. Will remove it.

signalfx/resource_signalfx_team_links.go
jrcamp: Just a dummy value to signal to TF that the state should be saved.

signalfx/resource_signalfx_team_links.go
jrcamp: Since it's a virtual resource I don't think this case really matters. We perform the same logic whether it's an update or create (calling SetId("") removes the state resulting in a create on next tf run)

keitwb (approved):

jrcamp:
Backend decided to keep previously deprecated method so it will be added back.

PR #267, opened 2020-11-13
is_default is always false for external and splunk link

dloucasfx:
Signed-off-by: Dani Louca <dlouca@splunk.com>

### DIFF
--- a/signalfx/resource_signalfx_data_link.go
+++ b/signalfx/resource_signalfx_data_link.go
@@ -67,12 +67,6 @@ func dataLinkResource() *schema.Resource {
 				Description: "Link to an external URL",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"is_default": &schema.Schema{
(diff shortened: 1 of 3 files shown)


signalfx/resource_signalfx_data_link.go
```diff
@@ -69,8 +69,7 @@ func dataLinkResource() *schema.Resource {
 					Schema: map[string]*schema.Schema{
 						"is_default": &schema.Schema{
 							Type:        schema.TypeBool,
-							Default:     true,
-							Optional:    true,
+							Computed:    true,
```
keitwb: What computes it?

dloucasfx: in the getPayloadDataLink, I hardcoded it to false for both external link and target_splunk 
The idea is not to make this field configurable , but we still need to set it in the struct, so it should be defined in the schema... Is there a different option to use?

jrcamp (approved):

PR #271, opened 2020-12-03
Added timezone argument to set Calendar Window Time Zone in the chart

rajeshkurup:


keitwb (approved):
@jrcamp Can you take a look as well and merge?

signalfx/resource_signalfx_single_value_chart.go
```diff
@@ -219,13 +225,16 @@ func getSingleValueChartOptions(d *schema.ResourceData) *chart.Options {
 	}
 
 	var programOptions *chart.GeneralOptions
+	if programOptions == nil {
```
jrcamp: This isn't doing anything since programOptions is always nil one line after the variable is declared. I think the intention was to have programOptions be nil unless some value in it was set.

So I think the correct thing to do is in the `if val, ok := d.GetOk("timezone"); ok {` block add

```go
if programOptions == nil {	
    programOptions = &chart.GeneralOptions{}	
}
```

rajeshkurup: Sorry for making silly mistake...
It was a copy-paste error.
fixed it now.

jrcamp:
I assume it was done like this for a reason, let's just keep the existing pattern so that programOptions stays nil unless an option needs to be set in it.

```go
	if val, ok := d.GetOk("max_delay"); ok {		
		if programOptions == nil {	
			programOptions = &chart.GeneralOptions{}	
		}	
		md := int32(val.(int) * 1000)			
		programOptions.MaxDelay = &md			
	}
```

Do the same condition in your code:

```go
		if programOptions == nil {	
			programOptions = &chart.GeneralOptions{}	
		}	
```

rajeshkurup:
@jrcamp Fixed your comments. thanks!

jrcamp (approved):

jrcamp:
Released as v6.2.0.

PR #283, opened 2021-01-22
Extending signalfx_detector with tags

apakulov-stripe:
Adding missing bits such as `tags` to `signalfx_detector` resource.

Both `POST` and `PUT` methods have a way to define `tags` for the detectors, but unfortunately it cant be managed via Terraform resource:
- https://dev.splunk.com/observability/reference/api/detectors/latest#endpoint-create-single-detector
- https://dev.splunk.com/observability/reference/api/detectors/latest#endpoint-update-single-detector

For the context, we're planning to use `tags` data in a search query: https://dev.splunk.com/observability/reference/api/detectors/latest#endpoint-retrieve-detectors-query

### DIFF
--- a/signalfx/resource_signalfx_detector.go
+++ b/signalfx/resource_signalfx_detector.go
@@ -89,6 +89,12 @@ func detectorResource() *schema.Resource {
 				Description:   "Seconds since epoch. Used for visualization",
 				ValidateFunc:  validation.IntAtLeast(0),
 			},
+			"tags": &schema.Schema{
+				Type:        schema.TypeList,
+				Optional:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "Tags associated with the detector",
+			},
 			"teams": &schema.Schema{
 				Type:        schema.TypeList,
 				Optional:    true,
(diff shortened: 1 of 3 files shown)


jrcamp:
lgtm, thanks for adding! Can you add it docs as well? https://github.com/splunk-terraform/terraform-provider-signalfx/blob/main/website/docs/r/detector.html.markdown

apakulov-stripe:
@jrcamp added in 7bc38b5

jrcamp:
Thanks! Do you need a release?

apakulov-stripe:
@jrcamp yep, it will be nice to have an actual release.
thanks in advance!

apakulov-stripe:
@jrcamp do you have any ETA on the upcoming release?

jrcamp:
Sorry, thanks for the reminder. Released v6.5.0.

signalfx/resource_signalfx_detector.go
```diff
@@ -321,6 +327,14 @@ func getPayloadDetector(d *schema.ResourceData) (*detector.CreateUpdateDetectorR
 	maxDelay := int32(d.Get("max_delay").(int) * 1000)
 	minDelay := int32(d.Get("min_delay").(int) * 1000)
 
+	var tags []string
+	if val, ok := d.GetOk("tags"); ok {
+		tags := []string{}
```
pdecat, line 332: This shadows the external `tags` variable, so tags are never actually set. See https://github.com/splunk-terraform/terraform-provider-signalfx/pull/306#issuecomment-870415564

PR #289, opened 2021-03-01
updated victorops integration resource name

tparikh:
The documentation for signalfx victorops integration was outdated. The resource should be `signalfx_victor_ops_integration` and not `signalfx_victor_ops_resource`. The current documentation & example will result in following error:

```
The provider provider.signalfx does not support resource type
"signalfx_victor_ops_resource".
```


jrcamp (approved):
Thanks!

PR #300, opened 2021-05-11
Fix AWS integration resource

asuresh4:
The AWS integration resource currently sets `default_action` to `Exclude` for `custom_namespace_sync_rule`. It should be left empty instead like for `namespace_sync_rule`.  This will ensure metrics are properly collected from custom namespaces when no explicit filters are provided, right now they're being excluded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased Changes
+
+BUGFIXES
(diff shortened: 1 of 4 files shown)


jrcamp (approved):

jrcamp:
Make sure you update CHANGELOG and any docs that need updating.

jrcamp (approved):

PR #312, opened 2021-08-25
fix changelog v6.7.6

xp-1000:


asuresh4:
Incorporated this update in #316

PR #319, opened 2021-09-27
resource/dashboardgroup: Handle updates with mirrors correctly

### LINKED ISSUE #264: Terraform apply isn't idempotent in scenario where dashboard_group with mirrors + dashboard resources are used
Hi there team,

I wanted to present a use case wherein I'm seeing some unexpected behavior and am wondering what you think about it. It's a little hard to see without an example, so I'll provide one after describing the scenario. In my configuration, I use a `resource signalfx_dashboard_group` and include dashboard mirrors via the `dashboard` argument (ref: https://registry.terraform.io/providers/splunk-terraform/signalfx/latest/docs/resources/dashboard_group#dashboard) AND create other `resource signalfx_dashboard` within the configuration to be added to the previously mentioned dashboard group. Running `terraform apply` results in a dashboard group being created with the mirrored dashboards + other dashboard resources, which is what I expect.

However, when I run `terraform apply` again, with no changes to the configuration, this time, I get a dashboard group with only the mirrored dashboards, and the other previously seen dashboard resources are gone, which is unexpected.

Here's an example. Let's say I have the following file structure:
```
/terraform
  -> dashboard_group/
     -> dashboard_group.tf
  -> dashboard_a/
      -> dashboard_a.tf
  -> dashboard_b/
      -> dashboard_b.tf
  -> main.tf
```

`terraform/dashboard_group/dashboard_group.tf` looks like this:
```
resource "signalfx_dashboard_group" "dashboardgroup0" {
  dashboard {
    dashboard_id = "mirror_A_Id"
  }

  dashboard {
    dashboard_id = "mirror_B_Id"
  }
}

output "dashboard_group_id" {
  value = signalfx_dashboard_group.dashboardgroup0.id
}
```

`terraform/main.tf` looks like this:
```
module "dashboard_group" {
  source       = "./dashboard_group"
}

module "dashboard_a" {
  source  = "./dashboard_a"
  dashboard_group_id = module.dashboard_group.dashboard_group_id
}

module "dashboard_b" {
  source  = "./dashboard_b"
  dashboard_group_id = module.dashboard_group.dashboard_group_id
}
```

I'll exclude what `terraform/dashboard_a.tf` and `terraform/dashboard_b.tf` look like but just note that they are each creating a `resource "signalfx_dashboard"` and specifying `var.dashboard_group_id` for the required property `dashboard_group` (ref: https://registry.terraform.io/providers/splunk-terraform/signalfx/latest/docs/resources/dashboard#argument-reference).

Steps:
1. `terraform apply` -> dashboardgroup0 contains a list of dashboards[mirror_A_Id, mirror_B_id, dashboard_a.id, dashboard_b.id]. This is expected.
2. `terraform apply` (note no changes in configuration at all) -> dashboardgroup0 contains list of dashboards [mirror_A_Id, mirror_B_id]. This is unexpected; why would dashboard_a and dashboard_b get removed?

I discovered this bug when I was using v4.18.0 but the problem is still present in v6.0.0 (latest version as of writing).

I would really appreciate your help here. I hope I've at least provided enough details to reproduce the behavior.

asuresh4:
The Terraform provider does not track non-mirrored dashboards in dashboard groups. It keeps track dashboardgroup membership on the respective dashboards. However, the backend modifies dashboard groups object to hold a list of all dashboards it holds, both mirrored and non-mirrored. The API expects this full list of dashboards whenever mirrored dashboards are added/present. This behavior is noted in step 4 of the API docs [here](https://dev.splunk.com/observability/docs/chartsdashboards/dashboard_groups_overview#Add-the-mirrored-dashboard). 

Currently, it does not seem possible to track info about dashboard membership on dashboard group resources with Terraform state as that will introduce a cyclic dependency. The API requires a dashboard group ID for dashboards, so the the dashboard group will need to be created before the member dashboards themselves. As a workaround, this PR attempts to fix the issue by doing a get on the dashboard group to collect all member dashboard ids and append it to list of dashboards with every update request. For this to work as expected it will also be required to set `-parallelism=1` .

Fixes #264

### DIFF
--- a/signalfx/resource_signalfx_dashboard_group.go
+++ b/signalfx/resource_signalfx_dashboard_group.go
@@ -518,8 +518,11 @@ func dashboardgroupUpdate(d *schema.ResourceData, meta interface{}) error {
 }
 
 func getNonMirroredDashes(config *signalfxConfig, d *schema.ResourceData) ([]*dashboard_group.DashboardConfig, error) {
-	oldDashboardList, newDashboardList := d.GetChange("dashboard")
-	mirrorIDsToBeOmitted := getMirrorsToBeOmitted(oldDashboardList, newDashboardList)
+	var mirrorIDsToBeOmitted map[string]bool
+	if d.HasChange("dashboard") {
+		oldDashboardList, newDashboardList := d.GetChange("dashboard")
+		mirrorIDsToBeOmitted = getMirrorsToBeOmitted(oldDashboardList, newDashboardList)
+	}
 
 	dg, err := config.Client.GetDashboardGroup(context.TODO(), d.Id())
 	if err != nil {
--- a/signalfx/resource_signalfx_dashboard_group.go
+++ b/signalfx/resource_signalfx_dashboard_group.go
@@ -3,6 +3,7 @@ package signalfx
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"log"
 	"strings"
 
@@ -404,12 +405,10 @@ func dashboardGroupAPIToTF(d *schema.ResourceData, dg *dashboard_group.Dashboard
 		dConfigs := make([]map[string]interface{}, len(dg.DashboardConfigs))
 
 		for i, dc := range dg.DashboardConfigs {
-			if dc.DescriptionOverride == "" && dc.NameOverride == "" && dc.FiltersOverride == nil {
-				// This is not a mirror, just a placeholder for a dashboard in the group
-				continue
-			} else {
-				// A real mirror, change the flag so we know to add it
+			if isMirroredDashboard(dc) {
 				hasMirrors = true
+			} else {
+				continue
 			}
 
 			dConf := make(map[string]interface{})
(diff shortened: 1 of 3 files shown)


dloucasfx:
@asuresh4 the fix LGTM , especially from the use case the bug is trying to address. 
my only concern with this approach is that we are not saving the the non-mirrored dashboard in the terraform dashboard group state ; would it create unwanted results if the group was updated outside the terraform? update/add/remove dashboards?

asuresh4:
Thanks for taking a look @dloucasfx ! Given that this update introduces a get of the dashboard group before an update, I believe the provider will end up with the latest dashboard group object from the backend, especially the list of dashboards since that's the only value we're interested in under the assumption one does not try to apply changes via Terraform and almost simultaneously make manual updates. In that case, there could be a race condition and the get call may not have the latest values. Were you able to think of any other scenario this could be problematic?

signalfx/resource_signalfx_dashboard_group.go
```diff
@@ -503,6 +517,59 @@ func dashboardgroupUpdate(d *schema.ResourceData, meta interface{}) error {
 	return dashboardGroupAPIToTF(d, dg)
 }
 
+func getNonMirroredDashes(config *signalfxConfig, d *schema.ResourceData) ([]*dashboard_group.DashboardConfig, error) {
+	oldDashboardList, newDashboardList := d.GetChange("dashboard")
```
dloucasfx:
````
// GetChange returns the old and new value for a given key.
//
// HasChange should be used to check if a change exists. It is possible
// that both the old and new value are the same if the old value was not
// set and the new value is. This is common, for example, for boolean
// fields which have a zero value of false.
````
Do you want to use `HasChange` to short circuit the call here?

dloucasfx (changes requested):
Other then this one comment about `HasChange`, the new logic to detect what mirrored dashboards looks solid!

signalfx/resource_signalfx_dashboard_group.go
asuresh4: Done

PR #328, opened 2021-11-23
update signalfx-go to v1.8.3 to support new cloud integrations services

xp-1000:
while the only way to configure all services on cloud integrations from terrraform is to use the data sources it needs to keep them up to date and release a new version of the provider to make them available to use.


xp-1000:
Hi @jrcamp @asuresh4 @jcheng-splunk @keitwb, can you merge/release soon ? configuring new services using terraform is not possible waiting for this. thanks

xp-1000:
conflict due to v6.7.9 release resolved

PR #329, opened 2021-11-30
teams should not impose ordering on dashboard_groups by using list

jherbage:
The use of typelist is causing our repo to constantly update dashboard groups as the teams order varies between environments.
Teams has no order so a set is the correct type to use surely.


keitwb (approved):
Looks good.

PR #331, opened 2021-12-03
Teams in dashboard groups  should be an unordered set rather than a list

jherbage:
I had previously missed off converting the set to a list in the get causing errors for anyone using the new provider


keitwb (approved):
Looks good.

PR #342, opened 2022-02-04
Fix and update org token

dloucasfx:
This PR fixes the `dpm_limits` , which a single [line](https://github.com/splunk-terraform/terraform-provider-signalfx/compare/SWAT-4236?expand=1#diff-6db680519113a56285c31717f8f0159ebf84939fdfc348fa8d4dfb3a32a84fc7R280) fix; but fixing this and/or running the existing test `TestAccCreateUpdateOrgToken` will fail with a server side error:
```
"code" : 500,
  "message" : "Server error occurred. Unique error identifier [24df50a1-e751-4a42-9adb-8eb759e875b9]"
```

This error is returned when the provider attempts to update an org token with a DPM/Category limit and not setting the `authScopes` in its payload.

In this PR I added `auth_scopes` to the token schema and I made it optional to keep it backward compatible;
Note, when you create a token, the server defaults to setting the value to `["API", "INGEST"]` so the initial creation succeeds whether you set `authScopes` or not; but due to the TF SDK not supporting `Default` [for complex types](https://github.com/hashicorp/terraform-plugin-sdk/issues/142) , in this case `TypeList`, I opted to set `Computed` to true; this will prevent false positive plan diffs .

I also added new test.

NOTE: ~~Do Not Merge as this depends on this PR https://github.com/signalfx/signalfx-go/pull/146 , once the it's merged, I need to update this PR to pull the new version of signalfx-go~~

update:  signalfx-go is updated

Signed-off-by: Dani Louca <dlouca@splunk.com>

### DIFF
--- a/signalfx/resource_signalfx_org_token.go
+++ b/signalfx/resource_signalfx_org_token.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"sort"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -23,6 +24,13 @@ func orgTokenResource() *schema.Resource {
 				Type:        schema.TypeString,
 				Optional:    true,
 				Description: "Description of the token (Optional)",
+			},
+			"auth_scopes": &schema.Schema{
+				Type:        schema.TypeList,
+				Optional:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Computed:    true,
+				Description: "Authentication scope, ex: INGEST, API, RUM ... (Optional)",
 			},
 			"disabled": &schema.Schema{
 				Type:        schema.TypeBool,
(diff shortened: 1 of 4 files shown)


signalfx/resource_signalfx_org_token.go
```diff
@@ -256,7 +277,7 @@ func orgTokenAPIToTF(d *schema.ResourceData, t *orgtoken.Token) error {
 			if limits.DpmNotificationThreshold != nil {
 				dpmStuff["dpm_notification_threshold"] = limits.DpmNotificationThreshold
 			}
-			if err := d.Set("dpm_limit", dpmStuff); err != nil {
+			if err := d.Set("dpm_limits", []map[string]interface{}{dpmStuff}); err != nil {
```
keitwb, line 280: What does this change have to do with the auth scope thing?

dloucasfx: the initial issue reported by the customer is that setting the limit on a token fails with this error:
```
signalfx_org_token.my_test_token: Creating...
╷
│ Error: Invalid address to set: []string{"dpm_limit"}
│
│ with signalfx_org_token.my_test_token,
│ on main.tf line 5, in resource "signalfx_org_token" "my_test_token":
│ 5: resource "signalfx_org_token" "my_test_token" {
```

I was able to reproduce right away and tbh I am not sure how this ever worked, this was broken right from the beginning .
I also noticed that the existing integration test was failing, due to the auth scope.

keitwb (approved):
Ok looks good.

PR #349, opened 2022-03-02
Add tags to the signalfx_dashboard resource

jordanbertasso:
This change allows tags to be specified for the signalfx_dashboard resource.

The existing functionality to specify tags for a dashboard [already existed in the signalfx-go library API](https://github.com/signalfx/signalfx-go/blob/master/dashboard/model_create_update_dashboard_request.go#L32).

### DIFF
--- a/signalfx/resource_signalfx_dashboard.go
+++ b/signalfx/resource_signalfx_dashboard.go
@@ -36,6 +36,12 @@ func dashboardResource() *schema.Resource {
 				Optional:    true,
 				Description: "Description of the dashboard (Optional)",
 			},
+			"tags": &schema.Schema{
+				Type:        schema.TypeList,
+				Optional:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: "Tags of the dashboard",
+			},
 			"charts_resolution": &schema.Schema{
 				Type:         schema.TypeString,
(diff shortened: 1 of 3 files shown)


signalfx/resource_signalfx_dashboard.go
```diff
@@ -36,6 +36,12 @@ func dashboardResource() *schema.Resource {
 				Optional:    true,
 				Description: "Description of the dashboard (Optional)",
 			},
+			"tags": &schema.Schema{
+				Type:        schema.TypeList,
```
jordanbertasso, line 40: Should this be a `schema.TypeSet`?

The only issue with making it a `schema.TypeSet` is I am not sure if I can check the values of the tags as done [here](https://github.com/splunk-terraform/terraform-provider-signalfx/pull/349/files#diff-b913ab62f26f9f3a350f1d6fffce493edf1d232cca5362e70c4a58161f564467R90).

leandropadua: since the api don't restrict tags to be unique, I think List is the right type here.

leandropadua (approved):
Looks good to me. Nice work.

bgola-signalfx:
Thanks for your contribution @jordanbertasso! 👍

PR #344, opened 2022-02-10
Add metric_streams_sync_state to signalfx_aws_integration

pdecat:
This PR add `use_metric_streams_sync` to `signalfx_aws_integration`.

Parly resolves #318 (does not address the log part).

TODO:
- [ ] rebase once https://github.com/splunk-terraform/terraform-provider-signalfx/pull/343 is merged
- [x] use new release of signalfx-go once https://github.com/signalfx/signalfx-go/pull/147 is merged and released
- [x] add documentation

PR #354, opened 2022-03-08
Refactor AWS integration

bgola-signalfx:
FYI  @pdecat

### DIFF
--- a/signalfx/resource_signalfx_aws_external_integration.go
+++ b/signalfx/resource_signalfx_aws_external_integration.go
@@ -1,13 +1,8 @@
 package signalfx
 
 import (
-	"context"
-	"encoding/json"
-	"fmt"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/signalfx/signalfx-go/integration"
-	"log"
-	"strings"
 )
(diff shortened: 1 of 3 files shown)


pdecat:
Looking good! 

Is that a first step towards merging `signalfx_aws_external_integration` and `signalfx_aws_token_integration` resources into `signalfx_aws_integration` as [I suggested in my other PR](https://github.com/splunk-terraform/terraform-provider-signalfx/pull/344#discussion_r808799334)?

bgola-signalfx:
> Is that a first step towards merging `signalfx_aws_external_integration` and `signalfx_aws_token_integration` resources into `signalfx_aws_integration` as [I suggested in my other PR](https://github.com/splunk-terraform/terraform-provider-signalfx/pull/344#discussion_r808799334)?

For now I do not plan to merge external and token resources as that would be a breaking change forcing customers to update their TF scripts. I agree it would be better to have just one resource type however, after this PR is merged, it should be fairly easy to maintain the AWS integration related code as it is the same code re-used in two places.

jkawecki-signalfx (approved):

akonarska (approved):

PR #359, opened 2022-03-15
Add additional_services and resource_filter_rules to Azure integration

bgola-signalfx:


signalfx/resource_signalfx_azure_integration_test.go
```diff
@@ -49,7 +49,20 @@ resource "signalfx_azure_integration" "azure_myteamXX" {
 
     services = [ "microsoft.sql/servers/elasticpools" ]
 
+	additional_services = [ "foo", "bar" ]
```
akonarska: nit: looks like formatting is wrong here

akonarska (approved):

signalfx/resource_signalfx_azure_integration_test.go
bgola-signalfx: fixed

PR #360, opened 2022-03-17
Fixes for Alert Muting Rule

luke-sterkowicz:
- Allowing fake delete of alert muting rules in the past (API limitation that causes invalid terraform state) - this was requested by customers
- Making 'filter' set optional - at least one filter is required by the API, however introducing the 'detectors' list makes the requirement for 'filter' set obsolete. For the cases where we have only 'detector' filter, terraform still would require 'filter' to be set. Workaround by placing the detector in 'filter' set is not perfect as it will cause conversions underneath and unnecessary updates on every terraform apply.


keitwb (approved):
ok, seems like a reasonable workaround.

PR #378, opened 2022-06-10
Change `rules` from Set to List in detector resource

bohdanborovskyi-ma:
We've faced a several issues when started migrating hundreds of existing detectors into Terraform due to `rule` being Set.

1. We can't ignore only changes in specific attributes of the `rule`, but rule block as a whole.
Taking into account that any threshold change introduced via SignalFX UI or API would diverge the `description` attribute of modified rule to some autogenerated text, it became very hard to keep our Terraform states with 0 drifts.
At the end we're used to completely ignore changes in the `rule` block, effectively loosing the possibility to modify other rule attributes from within Terraform code.

2. Another downside of TypeSet is that Terraform plan outputs always show re-creation of all rule objects instead of showing actual diff. 
For example, if I have 3 rules and wants to change the `disabled` attribute in one of them - TF plan would show that all old rules should be removed and 3 new ones would be added as it can't match old Set items with new ones.

P.s. I've built the provider locally & tested it on several existing detectors & internally developed modules without any compatibility issues noticed.
Please let me know if there're some technical constrains to merge this change

### DIFF
--- a/signalfx/resource_signalfx_detector.go
+++ b/signalfx/resource_signalfx_detector.go
@@ -109,9 +109,9 @@ func detectorResource() *schema.Resource {
 				Description: "Team IDs to associate the detector to",
 			},
 			"rule": &schema.Schema{
-				Type:        schema.TypeSet,
+				Type:        schema.TypeList,
 				Required:    true,
-				Description: "Set of rules used for alerting",
+				Description: "List of rules used for alerting",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"severity": &schema.Schema{
@@ -167,7 +167,7 @@ func detectorResource() *schema.Resource {
 						},
 					},
 				},
-				Set: resourceRuleHash,
+				//Set: resourceRuleHash,
 			},
 			"authorized_writer_teams": &schema.Schema{
 				Type:        schema.TypeSet,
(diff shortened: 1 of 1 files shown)


bohdanborovskyi-ma:
@keitwb @mbojko-signalfx Could you please share your feedback on proposed change?

Described downsides of `Set` type severely impact the flexibility & usage convenience for any module which creates detector resources under the hood

mbojko-signalfx:
Proposed change seems to be reasonable, I'll need to run few test to see if it doesn't introduce backward compatibility issues.

signalfx/resource_signalfx_detector.go
```diff
@@ -288,7 +288,7 @@ func timeRangeStateUpgradeV0(rawState map[string]interface{}, meta interface{})
 */
 func getPayloadDetector(d *schema.ResourceData) (*detector.CreateUpdateDetectorRequest, error) {
 
-	tfRules := d.Get("rule").(*schema.Set).List()
+	tfRules := d.Get("rule").([]interface{}).List()
```
mbojko-signalfx, line 291: There is no `List()` method on an interface slice

signalfx/resource_signalfx_detector.go
```diff
@@ -790,7 +790,7 @@ func validateProgramTextCondition(d *schema.ResourceDiff, meta interface{}) bool
 */
 func validateProgramText(d *schema.ResourceDiff, meta interface{}) error {
 
-	tfRules := d.Get("rule").(*schema.Set).List()
+	tfRules := d.Get("rule").([]interface{}).List()
```
mbojko-signalfx, line 793: There is no `List()` method on an interface slice

mbojko-signalfx:
I've had time to verify the proposed change and unfortunately I've noticed backward compatibility issues. 
Changing from `Set` to `List` introduces ordering into the `rules` collection, so it may result in unexpected changes in terraform plan even if rules were not modified. This is unfortunate limitation of the `terraform-plugin-sdk` at the moment.

mbojko-signalfx (changes requested):

atoulme:
Closing as this PR has gone stale. Please reopen and update if more work is appropriate.

PR #396, opened 2022-10-19
Improve the way dashboard group tells what mirrored dashboards are

seonsfx:
Dashboard group is telling if a dashboard is mirrored or not based on whether there is any override field or not in the dashboard config object.

This makes wrong judgements often times and skips adding mirrored dashboard that should have been added to the dashboard group.

Then, this misjudgement leads to the existing dashboard in the config not being recognized and being added over and over whenever the config is applied instead of being idempotent with the config.

This change also improves dashboard trackability by keeping track of config_id of dashboard configs.

### DIFF
--- a/signalfx/resource_signalfx_dashboard_group.go
+++ b/signalfx/resource_signalfx_dashboard_group.go
@@ -37,6 +37,11 @@ func dashboardGroupResource() *schema.Resource {
 				Description: "Dashboard IDs that are members of this dashboard group. Also handles 'mirrored' dashboards.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"config_id": &schema.Schema{
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "Unique identifier of an association between a dashboard group and a dashboard",
+						},
 						"dashboard_id": &schema.Schema{
 							Type:        schema.TypeString,
 							Required:    true,
@@ -422,10 +427,10 @@ func dashboardgroupCreate(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.SetId(dg.Id)
 
-	return dashboardGroupAPIToTF(d, dg)
+	return dashboardGroupAPIToTF(d, dg, meta)
 }
 
-func dashboardGroupAPIToTF(d *schema.ResourceData, dg *dashboard_group.DashboardGroup) error {
+func dashboardGroupAPIToTF(d *schema.ResourceData, dg *dashboard_group.DashboardGroup, meta interface{}) error {
 	debugOutput, _ := json.Marshal(dg)
 	log.Printf("[DEBUG] SignalFx: Got Dashboard Group to enState: %s", string(debugOutput))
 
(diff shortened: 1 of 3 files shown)


signalfx/resource_signalfx_dashboard_group.go
```diff
@@ -592,14 +598,16 @@ func dashboardgroupUpdate(d *schema.ResourceData, meta interface{}) error {
...
+	return dashboardGroupAPIToTF(d, dg, meta)
 }
 
 func getNonMirroredDashes(config *signalfxConfig, d *schema.ResourceData) ([]*dashboard_group.DashboardConfig, error) {
-	var mirrorIDsToBeOmitted map[string]bool
-	if d.HasChange("dashboard") {
-		oldDashboardList, newDashboardList := d.GetChange("dashboard")
-		mirrorIDsToBeOmitted = getMirrorsToBeOmitted(oldDashboardList, newDashboardList)
+	mirrorIDsToBeOmitted := map[string]bool{}
+	mirroredDashboardConfigs, err := getMirroredDashboardConfigs(config, d)
```
asuresh4, line 606: Is it safe to ignore errors returned by this method? It appears we only handle the happy path here. Would it make sense to log those errors?

signalfx/resource_signalfx_dashboard_group.go
```diff
@@ -37,6 +37,11 @@ func dashboardGroupResource() *schema.Resource {
 				Description: "Dashboard IDs that are members of this dashboard group. Also handles 'mirrored' dashboards.",
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"config_id": &schema.Schema{
```
asuresh4, line 40: I think we should also add a corresponding entry to the resource docs here: https://github.com/splunk-terraform/terraform-provider-signalfx/blob/main/website/docs/r/dashboard_group.html.markdown#argument-reference since we're adding a new resource field.

signalfx/resource_signalfx_dashboard_group.go
seonsfx: I thought it is okay to proceed with assuming all dashboards are non-mirror if we cannot fetch enough information. Now I think about it again, it should fail instead of proceeding because TF wii be tracking wrong info if we do so. I'll change it.

signalfx/resource_signalfx_dashboard_group.go
seonsfx: Great idea. I did not know we had such documentation within the repository. Since it is a computed field, I think it is more like an attribute than an argument. Just added it.

asuresh4:
LGTM!

asuresh4 (approved):

tiedotguy:
This is an incomplete fix.

I'm not 100% familiar with the TF SDK, so this might not be completely accurate, however as I understand it `dashboardGroupAPIToTF` fills out the provided `*schema.ResourceData`.  The issue is around [here](https://github.com/splunk-terraform/terraform-provider-signalfx/blob/main/signalfx/resource_signalfx_dashboard_group.go#L489):

```
		dConfigs := make([]map[string]interface{}, len(dg.DashboardConfigs))  // <------ This creates a slice of length N non-mirrors + M mirrored dashboards, all values being empty

		// Collect a list of mirrored dashboard configs
		config := meta.(*signalfxConfig)
		mirroredDashboardConfigs, err := getMirroredDashboardConfigs(config, d) // <----- This returns a slice containing M mirrored dashboards
		if err != nil {
			return err
		}

		for i, dc := range mirroredDashboardConfigs {
...
			dConfigs[i] = dConf // <----- This uses the index in to mirroredDashboardConfigs `i` as the index for `dConfigs`,  which is  invalid.
		}
		if len(mirroredDashboardConfigs) > 0 { 
			if err := d.Set("dashboard", dConfigs); err != nil { // <---- If there's any mirrors at all, store the slice that contains M mirrors and N **empty** definitions
...
```

The value for d.dashboard, if set, ends up with M valid mirror configurations and N **empty** dashboard definitions.  Those empty definitions are then resolved as needing to be deleted.

If the final line is replaced with:
```
			if err := d.Set("dashboard", dConfigs[:len(mirroredDashboardConfigs)]); err != nil {
```

It appears to solve the issue.  I don't have a super reliable reproduction though, so there may be some other problems that arise, such as ordering of items.

seonsfx:
@tiedotguy That is a great catch. There is a bug case because of the excessive dashboard being added to the state when a dashboard group has mirrored dashboard and non-mirrored dashboard at the same time. `len(dConfigs) - len(mirroredDashboardConfigs` equals to the number of non-mirrored dashboard in that case, and that number of blank dashboards are added to the tfstate. This results in Terraform recognizing some empty dashboards in the local state as a desired state update every time the same config is applied without any change. This does not appear to corrupt the original data stored in Splunk in this specific case though.

We will include the fix in the next release. Thank you very much for your observation and thorough analysis of the issue.

seonsfx:
PR's created: https://github.com/splunk-terraform/terraform-provider-signalfx/pull/413

PR #407, opened 2022-12-23
Improve state management of AWS integration

jkawecki-signalfx:


signalfx/resource_signalfx_aws_integration.go
```diff
@@ -325,9 +347,15 @@ func awsIntegrationAPIToTF(d *schema.ResourceData, aws *integration.AwsCloudWatc
 	if err := d.Set("use_metric_streams_sync", aws.MetricStreamsSyncState == "ENABLED"); err != nil {
 		return err
 	}
+	if err := d.Set("metric_streams_sync_state", strings.ToLower(aws.MetricStreamsSyncState)); err != nil {
```
penglerts, line 350: does it needs to be lowercase?

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -467,6 +495,28 @@ func getPayloadAWSIntegration(d *schema.ResourceData) (*integration.AwsCloudWatc
 		aws.LogsSyncState = "CANCELLING" // enable_logs_sync is false, and it has changed, meaning it was ENABLED before
 	}
 
+	if d.Get("metric_streams_sync_state") == "enabled" {
```
penglerts, line 498: it was quite complex for me to grasp, what about
```go
if d.HasChange("metric_streams_sync_state") {
  if d.Get("metric_streams_sync_state") == "enabled" {
    aws.MetricStreamsSyncState = "ENABLED"
 } else {
    // metric_streams_sync_state has changed, and it is not set to enabled, so it means it must be set to disabled, we need to cancell it then
      aws.MetricStreamsSyncState = "CANCELLING"
 }
} else {
   // state can have more values then TF model, so if it is no change we just passthru last recorded state
               old, _ := d.GetChange("metric_streams_sync_state")
		aws.MetricStreamsSyncState = strings.ToUpper(old.(string))
}

```

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -467,6 +495,28 @@ func getPayloadAWSIntegration(d *schema.ResourceData) (*integration.AwsCloudWatc
...
+	} else if d.HasChange("metric_streams_sync_state") {
+		// metric_streams_sync_state has changed, and it is not set to enabled, so it means it must be set to disabled
+		aws.MetricStreamsSyncState = "CANCELLING"
+	} else {
+		// it covers a case when changing configuration of integration for which Metric Streams sync is in CANCELLATION_FAILED state
+		old, _ := d.GetChange("metric_streams_sync_state")
+		aws.MetricStreamsSyncState = strings.ToUpper(old.(string))
+	}
+
+	if d.Get("logs_sync_state") == "enabled" {
```
penglerts, line 509: `logs` and `ms` have same logic so it can be moved to some common method

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -678,78 +757,38 @@ func DoIntegrationAWSDelete(d *schema.ResourceData, meta interface{}) error {
 	// Disable the AWS logs sync and/or CloudWatch metric streams sync if needed
 	needToDisableMetricStreams := int.Enabled && int.MetricStreamsSyncState != "" && int.MetricStreamsSyncState != "DISABLED"
 	needToDisableLogsSync := int.Enabled && int.LogsSyncState != "" && int.LogsSyncState != "DISABLED"
+	var toWaitFor []TargetState
 	if needToDisableMetricStreams || needToDisableLogsSync {
```
penglerts, line 761: this `if` seems redundant

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -678,78 +757,38 @@ func DoIntegrationAWSDelete(d *schema.ResourceData, meta interface{}) error {
...
 		if needToDisableMetricStreams {
 			int.MetricStreamsSyncState = "CANCELLING"
+			toWaitFor = append(toWaitFor, TargetState{"metric_streams_sync_state", "disabled"})
 		}
 		if needToDisableLogsSync {
 			int.LogsSyncState = "CANCELLING"
```
penglerts, line 767: i get confused when see `int` here and there :D  maybe it can be called `integration` ?

signalfx/resource_signalfx_aws_integration_waiter.go
```diff
@@ -0,0 +1,152 @@
...
+	},
+	"disabled": {
+		pending: []string{"ENABLED", "CANCELLING"},
+		target:  []string{"", "DISABLED"},
+	},
+}
+
+var stateTargetMapping = map[string]map[interface{}]stateTarget{
+	"enabled":                   integrationStateTargets,
+	"enable_logs_sync":          syncStateTargets,
```
penglerts, line 58: would it work given they are boolean?

penglerts (commented):
I think it will work fine for new parameters, but I am unsure what will happen when someone is using old boolean parameters... there are no Test Cases to verify  that...

on the contrary proposal I am thinking about:
- use only boolean params
- keep extra validation to enforce that logs and streams should be disabled when disabling integration
- let user disable integration even when we see cancelation_failed (threat this as disabled)
- on deletion check if all is disabled and prevent deletion unless all is disabled or flag `ignore_cancellation_failed` is set

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -678,78 +757,38 @@ func DoIntegrationAWSDelete(d *schema.ResourceData, meta interface{}) error {
...
-			int, err := config.Client.GetAWSCloudWatchIntegration(context.TODO(), intId)
-			if err != nil {
-				return 0, "", err
-			}
-			return int, stateSupplier(int), nil
-		},
-		Timeout:    d.Timeout(schema.TimeoutUpdate) - time.Minute,
-		Delay:      10 * time.Second,
-		MinTimeout: 5 * time.Second,
+	// Always disable integration before deleting it. This is crucial in case of usage of Metric Streams
```
penglerts, line 772: not sure if this is required ihmo it is enough to have logs and streams disabled

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -669,6 +736,18 @@ func integrationAWSUpdate(d *schema.ResourceData, meta interface{}) error {
...
+	// Makes sure integration is disabled. Not doing so may lead to an attempt of deleting
+	// an integration that is in DISABLING state.
+	if _, err := NewWaiter(d, config, d.Id()).WaitFor(TargetState{"enabled", false}); err != nil {
+		return err
+	}
+
+	return config.Client.DeleteAWSCloudWatchIntegration(context.TODO(), d.Id())
+}
+
+func integrationAWSDelete(d *schema.ResourceData, meta interface{}) error {
```
penglerts, line 748: why logic is now here and not in token/auth deletion method?

penglerts: ah ok... this is to have access to TF integration data?

README.md
```diff
@@ -29,7 +29,28 @@ $ make build
 
 Using the provider
 ----------------------
-If you're building the provider, follow the instructions to [install it as a plugin.](https://www.terraform.io/docs/plugins/basics.html#installing-a-plugin) After placing it into your plugins directory,  run `terraform init` to initialize it.
+
+If you want to test a local build of  the provider, follow the instructions to [Development Overrides for Provider Developers
```
bgola-signalfx, line 33: nit: follow the instructions on the following page: <link>

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -771,3 +810,49 @@ func validateAwsService(v interface{}, k string) (we []string, errors []error) {
...
+		if newValue == "disabled" && oldValue == "cancellation_failed" {
+			return fmt.Errorf("cancellation reattempt not allowed for %s", key)
+		}
+		return nil
+	}
+}
+
+func decorateWaitErrorOnUpdate(err error) error {
+	if strings.Contains(err.Error(), "CANCELLATION_FAILED") {
+		return fmt.Errorf("%w\nTBD manual troubleshot on update intructions go here", err)
```
bgola-signalfx, line 840: TBD :)

signalfx/resource_signalfx_aws_integration_state_mgnt_test.go
```diff
@@ -0,0 +1,169 @@
...
+resource "signalfx_aws_token_integration" "aws_myteam_token" {
+  name = "My AWS integration (tf dev)"
+}
+
+resource "signalfx_aws_integration" "aws_myteam" {
+  enabled = {{.Enabled}}
+
+  integration_id     = signalfx_aws_token_integration.aws_myteam_token.id
+  token              = "{{.AccessKeyID}}"
+  key			     = "{{.SecretAccessKey}}"
```
bgola-signalfx, line 22: nit: formatting

signalfx/resource_signalfx_aws_integration_state_mgnt_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+`
+
+type AwsIntegration struct {
+	Enabled bool
+
+	AccessKeyID     string
+	SecretAccessKey string
+
+	IgnoreCancelFailures *bool
```
bgola-signalfx, line 41: nit: `IgnoreCancellationFailure` (let's use the same name/wording everywhere)

signalfx/resource_signalfx_aws_integration_state_mgnt_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	MetricStreams *string
```
bgola-signalfx, line 43: nit: `MetricStreamsSyncState`

signalfx/resource_signalfx_aws_integration_state_mgnt_test.go
```diff
@@ -0,0 +1,169 @@
...
+	integration.IgnoreCancelFailures = &ignore
+	return integration
+}
+
+func (integration AwsIntegration) withMetricStreams(state string) AwsIntegration {
+	integration.MetricStreams = &state
+	return integration
+}
+
+func TestAccRejectCreatingOfDisabledIntegrationWithEnabledMetricStreams(t *testing.T) {
```
bgola-signalfx, line 62: Why? Seems to me this should pass via API (when user prepares a new integration but doesn't enable it yet).

signalfx/resource_signalfx_aws_integration_waiter.go
```diff
@@ -0,0 +1,152 @@
...
+
+	ignoreFailedCancel bool
+}
+
+type TargetState struct {
+	Name   string
+	Target interface{}
+}
+
+type stateTarget struct {
```
bgola-signalfx, line 29: nit: can we use different name here (TargetState vs. stateTarget)?

signalfx/resource_signalfx_aws_integration_waiter.go
```diff
@@ -0,0 +1,152 @@
...
+
+		timeout: d.Timeout(schema.TimeoutUpdate) - time.Minute,
+		delay:   10 * time.Second,
+
+		ignoreFailedCancel: ok && ignoreFailedCancel.(bool),
+	}
+}
+
+func (t *TargetStateWaiter) WaitForAll(props []TargetState) (int *integration.AwsCloudWatchIntegration, err error) {
+	for _, waitFor := range props {
```
bgola-signalfx, line 87: nit: local variable names: props --> states, waitFor --> state

PR #416, opened 2023-01-31
Add retryable http client

dscott-splunk:
Adds configurable retry logic

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -99,8 +99,6 @@ github.com/hashicorp/go-checkpoint v0.5.0/go.mod h1:7nfLNL10NsxqO4iWuW6tWW0HjZuD
 github.com/hashicorp/go-cleanhttp v0.5.0/go.mod h1:JpRdi6/HCYpAwUzNwuwqhbovhLtngrth3wmdIIUrZ80=
 github.com/hashicorp/go-cleanhttp v0.5.1 h1:dH3aiDG9Jvb5r5+bYHsikaOUIpcM0xvgMXVoDkXMzJM=
 github.com/hashicorp/go-cleanhttp v0.5.1/go.mod h1:JpRdi6/HCYpAwUzNwuwqhbovhLtngrth3wmdIIUrZ80=
-github.com/hashicorp/go-cleanhttp v0.5.2 h1:035FKYIWjmULyFRBKPs8TBQoi0x6d9G4xc9neXJWAZQ=
-github.com/hashicorp/go-cleanhttp v0.5.2/go.mod h1:kO/YDlP8L1346E6Sodw+PrpBSV4/SoxCXGY6BqNFT48=
 github.com/hashicorp/go-getter v1.4.0 h1:ENHNi8494porjD0ZhIrjlAHnveSFhY7hvOJrV/fsKkw=
 github.com/hashicorp/go-getter v1.4.0/go.mod h1:7qxyCd8rBfcShwsvxgIguu4KbS3l8bUCwg2Umn7RjeY=
 github.com/hashicorp/go-hclog v0.0.0-20180709165350-ff2cf002a8dd/go.mod h1:9bjs9uLqI8l75knNv3lV1kA55veR+WUPSiKIWcQHudI=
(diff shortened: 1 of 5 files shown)


signalfx/provider.go
```diff
@@ -153,27 +172,38 @@ func signalfxConfigure(data *schema.ResourceData) (interface{}, error) {
...
-	var netTransport = logging.NewTransport("SignalFx", &http.Transport{
-		Proxy: http.ProxyFromEnvironment,
-		DialContext: (&net.Dialer{
-			Timeout: 5 * time.Second,
-		}).DialContext,
-		TLSHandshakeTimeout: 5 * time.Second,
-		MaxIdleConns:        100,
-		MaxIdleConnsPerHost: 100,
-	})
+	pooledTransport := cleanhttp.DefaultPooledTransport()
```
dloucasfx: Is there a reason for using `cleanhttp`? I don't think we modify the transport object elsewhere in the provider, making the risk for race conditions nonexistent.

I don't have an issue using this package if we're trying to make the provider more resilient and there is a chance that we need it in the future; my only concern is keeping up with another package.

signalfx/provider.go
```diff
@@ -153,27 +172,38 @@ func signalfxConfigure(data *schema.ResourceData) (interface{}, error) {
...
+	pooledTransport.DialContext = (&net.Dialer{
+		Timeout: 5 * time.Second,
+	}).DialContext
+	pooledTransport.TLSHandshakeTimeout = 5 * time.Second
+	pooledTransport.MaxIdleConnsPerHost = 100
+
```
dloucasfx: If we are migrating to `cleanhttp` , let's set `MaxIdleConns` as before

dloucasfx (changes requested):
In general it looks good; please see my comments.
Thanks !

signalfx/provider.go
dscott-splunk: No reason, unfamiliar with go and ended up there thinking it was needed to get my changes to compile.  I agree with your thinking and removed `cleanhttp` 👍

signalfx/provider.go
dscott-splunk: Reverted back to using `net/http` and included `MaxIdleConns` again

go.sum
```diff
@@ -97,8 +97,9 @@ github.com/hashicorp/errwrap v1.0.0/go.mod h1:YH+1FKiLXxHSkmPseP+kNlulaMuP3n2brv
 github.com/hashicorp/go-checkpoint v0.5.0 h1:MFYpPZCnQqQTE18jFwSII6eUQrD/oxMFp3mlgcqk5mU=
 github.com/hashicorp/go-checkpoint v0.5.0/go.mod h1:7nfLNL10NsxqO4iWuW6tWW0HjZuDrwkBuEQsVcpCOgg=
 github.com/hashicorp/go-cleanhttp v0.5.0/go.mod h1:JpRdi6/HCYpAwUzNwuwqhbovhLtngrth3wmdIIUrZ80=
-github.com/hashicorp/go-cleanhttp v0.5.1 h1:dH3aiDG9Jvb5r5+bYHsikaOUIpcM0xvgMXVoDkXMzJM=
 github.com/hashicorp/go-cleanhttp v0.5.1/go.mod h1:JpRdi6/HCYpAwUzNwuwqhbovhLtngrth3wmdIIUrZ80=
+github.com/hashicorp/go-cleanhttp v0.5.2 h1:035FKYIWjmULyFRBKPs8TBQoi0x6d9G4xc9neXJWAZQ=
```
dloucasfx: I think this is a leftover from your initial change that you reverted, unless this indirectly needed by `go-retryablehttp`?
If not, I don't think there is any issue leaving it, the difference between the 2 versions is the addition of `ForceAttemptHTTP2` , but not sure if this will impact the SDK version we are on or not.
wdyt?

dscott-splunk: Good catch 👍 , probably safer to change as little as possible...I think I got it cleaned up now

dloucasfx (approved):

PR #434, opened 2023-06-17
fix: typo Gigibyte => Gibibyte has been corrected in the server side frontend

### LINKED ISSUE #433: Detector change on Splunk side : unit was Gigibyte and is now Gibibyte
Hello,

It seems Splunk has changed the configuration of the unit in the detectors, Gigibyte has now became Gibibyte (don't know why it was called gigibyte before :) )

If we create a detector with Gigibyte as the signal unit then the graph will not work in the Splunk interface (with errors in the browser console, regarding unit)
And indeed in the signal properties Gigibyte does not exist anymore and is replaced with Gibibyte

The official API documentation does not (yet) reflect this change, but this break all existing and future detectors. The API does not seem to check the value, so either Gigibyte or Gibibyte can be used but only Gibibyte will work in the web UI

Problem is trying to use the new corrected unit with this terraform module will fail 
`Error: expected viz_options.0.value_unit to be one of [Bit Kilobit Megabit Gigabit Terabit Petabit Exabit Zettabit Yottabit Byte Kibibyte Mebibyte Gigibyte Tebibyte Pebibyte Exbibyte Zebibyte Yobibyte Nanosecond Microsecond Millisecond Second Minute Hour Day Week], got Gibibyte`

So on Splunk side Gigibyte does not work anymore, and on the terraform module side we can't use its replacement...

pdecat:
Resolves #433 

I meant to add acceptance tests, but the API accepts both values without any validation.


signalfx/resource_signalfx_time_chart.go
```diff
@@ -1147,7 +1147,8 @@ var validateUnitTimeChart = validation.StringInSlice([]string{
 	"Byte",
 	"Kibibyte",
 	"Mebibyte",
-	"Gigibyte",
+	"Gibibyte",
+	"Gigibyte", // This is an historical typo that has been corrected on the server side frontend
```
pdecat: Should the old value be dropped to help existing users notice the change?

pdecat: An alternative could be to emit a warning if the value `Gigibyte` is passed: https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatefunc

pdecat:
Hi @benkeith-splunk, somehow PR reviews are still assigned to your other Github handle.

I guess this needs to be updated: https://github.com/splunk-terraform/terraform-provider-signalfx/blob/main/.github/CODEOWNERS

signalfx/resource_signalfx_time_chart.go
benkeith-splunk: We're about to do a major release so I'm ok with just removing it and making it a breaking change that we'll roll into the major release.

benkeith-splunk:
> Hi @benkeith-splunk, somehow PR reviews are still assigned to your other Github handle.
> 
> I guess this needs to be updated: https://github.com/splunk-terraform/terraform-provider-signalfx/blob/main/.github/CODEOWNERS

Thanks, fixed it.

signalfx/resource_signalfx_time_chart.go
pdecat: Done!

benkeith-splunk (approved):

benkeith-splunk:
Thanks for fixing this @pdecat . It should be available in the 7.0.0 release when we do it (hopefully later this week).

Issue #394, opened 2022-10-14
GovCloud Signalfx integrations


rss04:
When using this code with govcvloud, I am receiving the following error :

```
Error: Invalid address to set: []string{"external_id"}

```

Following code was used:
```
resource "signalfx_aws_integration" "aws_idemia_integration_st" {
  count          = var.use_security_token ? 1 : 0
  named_token    = signalfx_org_token.aws_integration.name
  integration_id = signalfx_aws_token_integration.aws_token.id

  #if auth_method = \"SecurityToken\"
  token = aws_iam_access_key.splunk_obs_iam_user_access_key[0].id
  key   = aws_iam_access_key.splunk_obs_iam_user_access_key[0].secret


  enable_aws_usage          = var.enable_aws_usage
  enable_check_large_volume = var.enable_check_large_volume
  enable_logs_sync          = var.enable_logs_sync
  enabled                   = var.enable_integration #defaults is true

  #custom_cloudwatch_namespaces = var.custom_cloudwatch_namespaces
  dynamic "custom_namespace_sync_rule" {
    for_each = var.custom_namespace_sync_rules
    content {
      default_action = try(custom_namespace_sync_rules.value["default_action"], "Exclude")
      filter_action  = try(custom_namespace_sync_rules.value["filter_action"], "Include")
      filter_source  = try(custom_namespace_sync_rules.value["filter_source"], "filter('code', '200')")
      namespace      = try(var.custom_namespace_sync_rules.value["namespace"], "Idemia-custom-namespace")
    }
  }
  sync_custom_namespaces_only = var.sync_custom_namespaces_only #defaults to false
  import_cloud_watch          = var.import_cloud_watch          #default is true
  dynamic "metric_stats_to_sync" {
    for_each = var.metric_stats_to_sync
    content {
      metric    = try(metric_stats_to_sync.value["metric"], "NetworkPacketsIn")
      namespace = try(metric_stats_to_sync.value["namespace"], "Idemia-custom-namespace")
      stats     = try(metric_stats_to_sync.value["stats"], ["upper"])
    }
  }
  #name        = var.integration_name

  dynamic "namespace_sync_rule" {
    for_each = var.monitor_aws_services ? [] : var.namespace_sync_rules
    content {
      default_action = try(custom_namespace_sync_rules.value["default_action"], "Exclude")
      filter_action  = try(custom_namespace_sync_rules.value["filter_action"], "Include")
[...]

PR #445, opened 2023-07-13
Fix external-id when token-based AWS integration is created in two steps

bgola-signalfx:
This change fixes an issue when signalfx_aws_token_integration and signalfx_aws_integration resources were created in two separate Terraform runs.

The issue was "panic: Invalid address to set: []string{"external_id"}".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ## 8.0.0
 IMPROVEMENTS:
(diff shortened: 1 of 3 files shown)


akonarska (approved):

akonarska:
Fixes https://github.com/splunk-terraform/terraform-provider-signalfx/issues/394

PR #450, opened 2023-08-16
Don't allow empty regions in AWS integration

anosek-an:


akonarska:
Please change also docs that are available here https://github.com/splunk-terraform/terraform-provider-signalfx/blob/main/website/docs/r/aws_integration.html.markdown

akonarska (approved):

PR #477, opened 2024-05-14
Add dwalasek as codeowner of SLO files

atoulme:
@dwalasek please review.


atoulme:
@dwalasek approved outside Github.

PR #490, opened 2024-07-11
detector: add an option to specify detector's origin and AutoDetect parent detector id for customized AutoDetect detector

dwalasek:
detector: add an option to specify detector's origin and AutoDetect parent detector id for customized AutoDetect detector


mbojko-signalfx (approved):
Looks good to me

PR #501, opened 2024-09-16
Update metric_ruleset.html.markdown

echoi-splk:
Fix html formatting of restoration

### DIFF
--- a/website/docs/r/metric_ruleset.html.markdown
+++ b/website/docs/r/metric_ruleset.html.markdown
@@ -3,7 +3,7 @@ layout: "signalfx"
 page_title: "Splunk Observability Cloud: signalfx_metric_ruleset"
 sidebar_current: "docs-signalfx-resource-metric-ruleset"
 description: |-
-Allows Terraform to create and manage Splunk Infrastructure Monitoring metric rulesets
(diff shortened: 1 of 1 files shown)


website/docs/r/metric_ruleset.html.markdown
```diff
@@ -94,4 +94,4 @@ The following arguments are supported in the resource block:
     * `stop_time` - (Optional) Time to which the restoration job will restore archived data, in the form of *nix time in milliseconds
 
 * `routing_rule` - (Required) Routing Rule object
-  * `destination` - (Required) - end destination of the input metric. Must be `RealTime`, `Archived`, or `Drop`
\ No newline at end of file
+  * `destination` - (Required) - end destination of the input metric. Must be `RealTime`, `Archived`, or `Drop`
```
kevalds51, line 97: Is this also causing the start/stop time to not be nested under restoration as well? 
---
Screenshot 2024-09-16 at 10 45 43 AM

kevalds51: We confirmed that the formatting is correct on our end and seems like an issue on the TF portal side

kevalds51 (approved):

PR #506, opened 2024-10-04
Feature: Add new package providermeta

MovieStoreGuy:
## Context

Extends: https://github.com/splunk-terraform/terraform-provider-signalfx/pull/505

I wanted an easier way to provide the following:

- Keep the existing behaviour for configuring the provider
- Make it easier to test
- Make it easier to add unit tests with locally mocked endpoints and functionality.

The new package moves all the logic that currently exists in `signalfx/provider.go` into the new package which makes it possible use within unit tests without causing circular import issues.

## Changes 

- Moved Provider configuration into its own type `MetaProviderLookup`
  - These functions are okay to fail since they were originally intended to inject the required provider information without hard coding these into the tests.
 - Added lookup functions for `FileMetaLookup`, `UserMetaLookup`, and `NetrcMetaLookup` to align with the existing configuration process.
 - Helper functions `LoadClient` and `LoadApplicationURL` to reduce repeated code across the different resources.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: "CLA Assistant"
         if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
-        uses: cla-assistant/github-action@v2.4.0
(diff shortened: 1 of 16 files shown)


atoulme:
~Please resolve conflicts.~ never mind, I merged the conflicts.

atoulme (approved):

PR #507, opened 2024-10-04
Feature: Adding package `tftest`

MovieStoreGuy:
## Context

Extends: https://github.com/splunk-terraform/terraform-provider-signalfx/pull/506

I wanted an easier way to validate the CRUD operations of a resource and the terraform provided package has a lot of expectations around setting up a lot of the terraform configuration around it. 

This allows for an easier framework to validate the CRUD operations of a resource and inject a `providermeta.Meta` with a client pointing to a local test server.

## Changes 

- Added `ResourceOperationTestCase` which allows for abstraction of testing the resource
  - It does also require setting up the Encoder and Decoder methods or they could be anon funcs as done inside the test files.
- Added `NewTestHTTPMetaMock` which allows for setting a dedicated http test server per test case to mock out some of the different conditions that could happen with the endpoints.


internal/check/timezone.go
```diff
@@ -8,7 +8,6 @@ import (
 	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-
```
atoulme, line 11: is this change needed?

atoulme (approved):

internal/check/timezone.go
MovieStoreGuy: Strictly not, it was an artifact of my all my rebasing.

PR #509, opened 2024-10-08
Feature: Creating dedicated resource definition for `sfx_team`

MovieStoreGuy:
## Context

The current library makes it hard to effectively isolate the required changes which introduces:

- Lack of standards (since each method needs to be a new name due to symbol collisions)
- Not easy to repeat practices
- Hard to effectively test (unit, and acceptance)

Moving each resource to own package helps address some of the above concerns and allows in future to make it easier to include new functionality if required.

The expected package layout is the following:

```
internal/definition
└── <api>
    ├── [resource,data]{,_test,_acceptance_test}.go
    ├── schema{,_test}.go
    └── testdata
        └── <example>.tf       
```

## Changes 

- Adds new package for `team` within definitions


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/splunk-terraform/terraform-provider-signalfx
 
-go 1.21
+go 1.23
```
MovieStoreGuy, line 3: Annoyingly required to have `http.NewServerMux` work as intended

atoulme (approved):

PR #515, opened 2024-10-09
Chore: Improving local lint support

MovieStoreGuy:
## Context

This looks to move a lot of the checks and formatting that is done in parts to use golangci-lint by default.

## Changes

- Removed deprecated script functionality
- Added `lint` and `lint-fix` into makefile
- Updated `build` to run `go build` instead of `go install`


GNUmakefile
```diff
@@ -52,13 +53,19 @@ checklicense: $(ADDLICENCESE)
...
-test: fmtcheck
-	go test -i $(TEST) || exit 1
-	echo $(TEST) | \
-		xargs -t -n4 go test $(TESTARGS) -timeout=30s -parallel=4
+.PHONY: lint-fix
+lint-fix:
+	$(GOLANGCI_LINT) run -v --fix
+
+build:
+	go build
```
MovieStoreGuy, line 65: I suspect that this will impact someone's workflow however, I am curious if it is used.

atoulme (approved):

PR #518, opened 2024-10-10
Deprecate: Removing original definitions for `signalfx/notification`

MovieStoreGuy:
## Context

This functionality has been ported and updated to be within the new packages providing this functionality, this closes the loop on the deprecation of the current definitions being used.

## Changes

- Removing `signalfx/notification` files
- Moved functionality to use the ported values in `check` and `common`


.golangci.yml
```diff
@@ -19,7 +19,7 @@ run:
   exclude-dirs:
     # This package is what is being migrated across to `internal/definition` 
     # so ignoring existing issues that exist and 
-    - signalfx
+    - (^|/)signalfx($|/)
```
atoulme: is this related?

atoulme (approved):

PR #523, opened 2024-10-14
Enhancement: Adopting `defintion.team` in main provider

MovieStoreGuy:
## Context

Extends: https://github.com/splunk-terraform/terraform-provider-signalfx/pull/522

On the pathway of migrating functionality to isolated packages, this works as a transition period. All of team's functionality has been moved to its own package and using type aliasing, It will continue to work as expected with the correct types. 

To help with the transition period, I have aliased the meta value used and the functions expected to create the schema for team.

## Changes

- Aliased `pmeta.Meta` to be `signalfx.signalfxConfig`, they are like for like so it is intended to work as expected.
- Aliased `teamResource` to be equal to `team.NewResource`


atoulme:
Please resolve the conflict

atoulme (approved):

PR #538, opened 2024-10-24
Feature: Improved detector validation

MovieStoreGuy:
## Context

This brings over the context from the existing detector solution and making sure there is coverage for it.

## Changes

- Added detector acceptance tests
- Added detector validation
- Added cleanup on missing detector


internal/common/notification_list.go
```diff
@@ -11,6 +11,7 @@ func NewNotificationList(items []any) ([]*notification.Notification, error) {
 	if len(items) == 0 {
 		return nil, nil
 	}
+
```
atoulme:
```suggestion
```

atoulme (approved):

PR #545, opened 2024-11-13
Feature: Add support for "Service Accounts"

MovieStoreGuy:
## Context

Since actions that require "admin" are only applied to session tokens, it means that customers need a means of supplying an session token. This can result in a session token that is stored within a vault service or user input for that given CI run.

This is typically fine in most cases considering admins have a lot of power in the account, however, in the scenario that you want to provide a greater level of automation, supplying a session token becomes cumbersome.

Hence, the idea of a "service account" then allows that increased level of autonomy since a session token can be create on behalf of the service account in the provider.

A service account is purely a user that can authenticate via email/password (meaning it has to be manually added into the org, and can't use an SSO user). This removes the 30d restriction that is applied to session tokens, and any external management that would be needed to be done by the user prior.

## Changes

- Provider supports authentication by "service accounts"
- Included documentation on what is meant by "service account"

## Intended Release
v9.3.0

No breaking changes since this currently extends the current configuration and doesn't change the existing behaviour.

### DIFF
--- a/internal/providermeta/meta.go
+++ b/internal/providermeta/meta.go
@@ -99,11 +99,11 @@ func (m *Meta) LoadSessionToken(ctx context.Context) (string, error) {
 	return resp.AccessToken, nil
 }
 
-func (s *Meta) Validate() (errs error) {
-	if s.AuthToken == "" && (s.Email == "" && s.Password == "") {
+func (m *Meta) Validate() (errs error) {
+	if m.AuthToken == "" && (m.Email == "" && m.Password == "") {
 		errs = multierr.Append(errs, errors.New("missing auth token or email and password"))
 	}
-	if s.APIURL == "" {
+	if m.APIURL == "" {
 		errs = multierr.Append(errs, errors.New("api url is not set"))
 	}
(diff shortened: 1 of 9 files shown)


internal/providermeta/meta.go
```diff
@@ -68,11 +72,38 @@ func LoadApplicationURL(ctx context.Context, meta any, fragments ...string) stri
...
+	}
+
+	// TODO: determine if any additional fields would be useful for debugging.
+	tflog.Info(ctx, "Created new session token")
+
+	return resp.AccessToken, nil
+}
+
+func (m *Meta) Validate() (errs error) {
+	if m.AuthToken == "" && (m.Email == "" && m.Password == "") {
```
mattpatl: If either email OR password are missing, credentials are incomplete so validation should fail.
`if m.AuthToken == "" && (m.Email == "" || m.Password == "")`

MovieStoreGuy: Included some test cases to cover this case as well.

atoulme (approved):

PR #549, opened 2024-11-20
SLO chart type support

lukasz-picur-signalfx:


lukasz-picur-signalfx:

lukasz-picur-signalfx:
recheck

signalfx/resource_signalfx_slo_chart.go
```diff
@@ -0,0 +1,122 @@
...
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "URL of the chart",
+			},
+		},
+
+		Create: slochartCreate,
+		Read:   slochartRead,
+		Update: slochartUpdate,
+		Delete: slochartDelete,
```
MovieStoreGuy: Please us the `<Operation>Context` methods since these methods are deprecated by the API.

signalfx/resource_signalfx_slo_chart.go
```diff
@@ -0,0 +1,122 @@
...
+		Exists: chartExists,
```
MovieStoreGuy: The API mentions that this is deprecated and not required if `Read` is implemented.

signalfx/resource_signalfx_slo_chart.go
```diff
@@ -0,0 +1,122 @@
...
+		SloId: d.Get("slo_id").(string),
+	}
+}
+
+func slochartCreate(d *schema.ResourceData, meta any) error {
+	config := meta.(*signalfxConfig)
+	payload := getPayloadSloChart(d)
+
+	debugOutput, _ := json.Marshal(payload)
+	log.Printf("[DEBUG] SignalFx: Create SLO Chart Payload: %s", string(debugOutput))
```
MovieStoreGuy: For this, you could do:

```go
tflog.Debug(ctx, "Created SLO chart", tfext.NewLogFields().JSON("payload", payload))
```

This would keep the code more concise.

signalfx/resource_signalfx_slo_chart.go
```diff
@@ -0,0 +1,122 @@
...
+
+	sloChart, err := config.Client.CreateSloChart(context.TODO(), payload)
+	if err != nil {
+		return err
+	}
+	// Since things worked, set the URL and move on
+	appURL, err := buildAppURL(config.CustomAppURL, CHART_APP_PATH+sloChart.Id)
```
MovieStoreGuy: You can use the following method to make sure this uri is correctly built.

```go
pmeta.LoadApplicationURL(ctx, meta, ChartAppPath, slochart.Id)
```

signalfx/resource_signalfx_slo_chart.go
```diff
@@ -0,0 +1,122 @@
...
+
+func slochartAPIToTF(d *schema.ResourceData, c *chart.Chart) error {
+	debugOutput, _ := json.Marshal(c)
+	log.Printf("[DEBUG] SignalFx: Got SLO Chart to enState: %s", string(debugOutput))
+
+	if err := d.Set("slo_id", c.SloId); err != nil {
+		return err
+	}
+
+	return nil
```
MovieStoreGuy: This can just be:

```suggestion
return d.Set("slo_id", c.SloId)
```

signalfx/resource_signalfx_slo_chart.go
```diff
@@ -0,0 +1,122 @@
...
+	if err := d.Set("url", appURL); err != nil {
+		return err
+	}
+	d.SetId(sloChart.Id)
+	return slochartAPIToTF(d, sloChart)
+}
+
+func slochartAPIToTF(d *schema.ResourceData, c *chart.Chart) error {
+	debugOutput, _ := json.Marshal(c)
+	log.Printf("[DEBUG] SignalFx: Got SLO Chart to enState: %s", string(debugOutput))
```
MovieStoreGuy: Considering there is only one field being set, I don't believe this is needed.

MovieStoreGuy (commented):
I've added comments where i think things could be improved, mostly as nits.

However, I do think adopting the Operation Context methods should be done since it they also provide the ability to gracefully stop resources being created/destroyed during the provider shutdown.

MovieStoreGuy (approved):
I suggest doing the fixes in a follow up PR thinking about it.

lukasz-picur-signalfx:
All of the suggestions are addressed.

MovieStoreGuy (approved):

PR #554, opened 2025-01-10
Add validation of viz objects

klisiecka-splunk:
Adding validation for dashboards, dashboard groups and charts: terraform provider should fail on `plan` if the payload is faulty.


signalfx/resource_signalfx_dashboard.go
```diff
@@ -812,6 +813,16 @@ func getDashboardFilters(d *schema.ResourceData) []*dashboard.ChartsSingleFilter
 	return filterList
 }
 
+func dashboardValidate(ctx context.Context, dashboardObject *schema.ResourceDiff, config interface{}) error {
```
MovieStoreGuy:
```suggestion
func dashboardValidate(ctx context.Context, dashboardObject *schema.ResourceDiff, config any) error {
```

signalfx/util.go
```diff
@@ -50,6 +50,22 @@ var ChartColorsSlice = []chartColor{
 	{"lime_green", "#6bd37e"},
 }
 
+// getPayload for viz objects is called withs: *schema.ResourceDiff or *schema.ResourceData - that's why we need this interface
+type ResourceDataAccess interface {
+	Get(key string) interface{}
+	GetOk(key string) (interface{}, bool)
+}
+
```
MovieStoreGuy: I would prefer to not add an interface within the utility file here, it will make it hard to maintain since it is somewhat hidden within the file of all the things that don't exactly fit within a resource object.

To be clear, the interface is fine but move it to a seperate file to make it easier to maintain and discover later on. 

Moreover, Go doc style should take the form of:

```golang
// <Definition> ...
<Definition>
```

Please update to be reflect the go doc style.

signalfx/util.go
```diff
@@ -50,6 +50,22 @@ var ChartColorsSlice = []chartColor{
...
+func chartValidate(ctx context.Context, chartResource *schema.ResourceDiff, config interface{}, getPayload func(ResourceDataAccess) (*chart.CreateUpdateChartRequest, error)) error {
+	payload, err := getPayload(chartResource)
+
+	if err != nil {
+		return err
+	}
+
+	return config.(*signalfxConfig).Client.ValidateChart(ctx, payload)
```
MovieStoreGuy: One thing to be conscious of is that if the type isn't the expected value, this code will panic, not expecting that it will but if it were to, it is a terrible user experience. 

I would suggest looking at type checking here.

signalfx/util.go
```diff
@@ -50,6 +50,22 @@ var ChartColorsSlice = []chartColor{
 	{"lime_green", "#6bd37e"},
 }
 
+// getPayload for viz objects is called withs: *schema.ResourceDiff or *schema.ResourceData - that's why we need this interface
+type ResourceDataAccess interface {
+	Get(key string) interface{}
+	GetOk(key string) (interface{}, bool)
+}
+
+func chartValidate(ctx context.Context, chartResource *schema.ResourceDiff, config interface{}, getPayload func(ResourceDataAccess) (*chart.CreateUpdateChartRequest, error)) error {
```
MovieStoreGuy: I wonder if it is more worthwhile to abstract this further instead of passing in placed type params. 

Something like:

```go 
type ChartValidatorFunc func(input) (ChartRequest, error)

func (fn ChartValidatorFunc) Validate(ctx, resource, config) error {
  if fn == nil {
   return errors.New("no data parser provided")
  }
  payload, err := fn(resource)
  if err != nil {
    return err
  }
  return config.ValidateChart(ctx, payload)
}
```

This is something that the standard lib does with `http.HandlerFunc` as a more concrete example.

MovieStoreGuy (changes requested):
Please ensure that test are also included with changes. This is important to ensure behaviour is kept between versions and error cases are correctly handled.

klisiecka-splunk:
I've added tests and implemented the suggested changes. Please let me know if there are any areas that need further improvement.

MovieStoreGuy:
Please make sure you sign the CLA.

klisiecka-splunk:

MovieStoreGuy (approved):

PR #558, opened 2025-01-11
Feature: AppDynamics data link

chrjosep:
## Feature
- In order to further integrate AppDynamics with Splunk o11y, we are adding data links that deep link to AppDynamics CSaaS.
- Add `target_appd_url` to data link resource
- Bump `signalfx-go` version

## Usage
- Run a Terraform file using the new `target_appd_url` field on a data link resource.

## Implementation
- Conditional check on data link type. Followed style for existing data link types (`target_external_url`, `target_splunk`, etc.).
- URL validation checks for `target_appd_url`
- Error handling for URL validation. Returns an error if input URL does not match AppDynamics regex pattern. 
  - URL should contain `.saas.appdynamics.com`, query values `component` and `application`. The query values should be positive integers

## Testing
- Manually tested by generating local binary and running Terraform module with `development_overrides` flag. 
- Acceptance tests:
  - Valid `target_appd_url` url that should create and destroy a data link without errors
  - A `target_appd_url` with a missing query parameter `component`. This should throw an error in the URL validation logic.

### DIFF
--- a/signalfx/resource_signalfx_data_link.go
+++ b/signalfx/resource_signalfx_data_link.go
@@ -246,7 +246,7 @@ func getPayloadDataLink(d *schema.ResourceData) (*datalink.CreateUpdateDataLinkR
 			}
 			match := re.MatchString(dl.URL)
 			if !match {
-				return dataLink, fmt.Errorf("Enter a valid AppD Link. The link needs to include the contoller URL, application ID, and Application component.")
+				return dataLink, fmt.Errorf("enter a valid AppD Link. The link needs to include the contoller URL, application ID, and Application component")
 			}
 
 			dataLink.Targets = append(dataLink.Targets, dl)
@@ -389,7 +389,7 @@ func dataLinkAPIToTF(d *schema.ResourceData, dl *datalink.DataLink) error {
 			return err
 		}
 	}
-	if appdLinks != nil && len(appdLinks) > 0 {
+	if len(appdLinks) > 0 {
 		if err := d.Set("target_appd_url", appdLinks); err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


chrjosep:

signalfx/resource_signalfx_data_link.go
```diff
@@ -206,6 +226,29 @@ func getPayloadDataLink(d *schema.ResourceData) (*datalink.CreateUpdateDataLinkR
 		}
 	}
 
+	if val, ok := d.GetOk("target_appd_url"); ok {
+		appdURLs := val.(*schema.Set).List()
+
+		appdUrlPatternRegex := "^https?:\\/\\/[a-zA-Z0-9-]+\\.saas\\.appdynamics\\.com\\/.*application=\\d+.*component=\\d+.*"
+		re := regexp.MustCompile(appdUrlPatternRegex)
```
MovieStoreGuy: Avoid using `MustCompile` or `Must*` methods in general since on error they will panic causing state to be incorrectly set if this was to be in the middle in an apply.

Be sure to use `Compile` instead and return the error.

MovieStoreGuy: Do you mind expanding on what the regular expression is testing?

Reading it it looks like you can have any wildcard for `*.saas.appdynamics.com` and expecting to have the query values for `application`, and `component`?

Did I miss anything here?

MovieStoreGuy (commented):
Just a comment from me.

signalfx/resource_signalfx_data_link.go
chrjosep: Yes that is correct. The other condition would be that the query values for `application` and `component` need to be positive integers.

MovieStoreGuy (approved):

PR #569, opened 2025-02-19
feature: Add aws integration definition

MovieStoreGuy:
## Context

Starting to move the aws integration into its own package to make it easier to maintain.

## Changes

- Adding awsintegration schema definition with decode.


internal/definition/awsintegration/schema_integration.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	if v, ok := rd.GetOk("enable_log_sync"); ok && v.(bool) {
+		cwi.LogsSyncState = "ENABLED"
+	} else if rd.HasChange("enable_log_sync") {
+		// enable_logs_sync is false, and it has changed, meaning it was ENABLED before
+		cwi.LogsSyncState = "CANCELING"
+	}
+
+	if v, ok := rd.GetOk("external_id"); ok && v != "" {
```
bgola-signalfx, line 270: I am not sure about this as on the API level `externalId` is always defined. Is it `nil` when `signalfx_aws_token_integration` resource is used? 

Shouldn't this logic depend simply on the `authMethod` field?

internal/definition/awsintegration/schema_integration.go
```diff
@@ -0,0 +1,312 @@
...
+		cwi.RoleArn = rd.Get("role_arn").(string)
+	} else if v, ok := rd.GetOk("token"); ok && v != "" {
+		cwi.AuthMethod = integration.SECURITY_TOKEN
+		cwi.Token = v.(string)
+		cwi.Key = rd.Get("key").(string)
+	} else {
+		return nil, fmt.Errorf("requires either `external_id` or `token` and `key`")
+	}
+
+	if len(cwi.Regions) == 0 {
```
bgola-signalfx: Is this more strict than before? I may be wrong but it seems to me that previously TF barked only when `services` were not defined at all but empty list got accepted.

internal/definition/awsintegration/schema_integration.go
MovieStoreGuy: That is a great point, let me clean that up.

MovieStoreGuy: >  Is it nil when signalfx_aws_token_integration resource is used?

Yes, if you have created the details for token integration. 

> Shouldn't this logic depend simply on the authMethod field?

So interestingly, this is currently a computed field meaning it isn't possible to be set by the user and so the provider figures it out.

internal/definition/awsintegration/schema_integration.go
MovieStoreGuy: It looks like the original code path would never actually execute, so this is technically is stricter in the sense that it will now validate that no values are part of the set.

internal/definition/awsintegration/schema_integration.go
bgola-signalfx: TBH I would rather have this logic depend on the `authMethod` if possible to avoid some regression in the future should the `externalId` become always defined (after all this is how it looks like on the API level).

MovieStoreGuy: Alright, I have updated the code to require users to set AuthMethod as part of their terraform config, and provide validation on it.

internal/definition/awsintegration/schema_integration.go
```diff
@@ -34,9 +34,16 @@ func newIntegrationSchema() map[string]*schema.Schema {
...
+			Type:     schema.TypeString,
+			Required: true,
+			ValidateFunc: validation.StringInSlice(
+				[]string{
+					string(integration.EXTERNAL_ID),
+					string(integration.SECURITY_TOKEN),
+				},
+				false,
+			),
+			Description: "The mechanism used to authenticate with AWS. Must be either `ExternalId` or `SecurityToken`.",
```
bgola-signalfx: Is this really necessary? Can't we leave the field spec as-is? 

Otherwise, if I understand correctly, all TF users would need to update their TF scripts?

MovieStoreGuy: Yeah, so I had made this change to make it explicit on what the auth provider is that is being used. 

This is mostly to address the original comment on using auth method as the means of figuring out what settings to use. This would be required in the create case since this field is currently computed so a user can not currently set it; it would require the most recent commit to be reverted.

bgola-signalfx (approved):

PR #573, opened 2025-03-13
[feature]: Adding support for managed metric streams externally

MovieStoreGuy:
## Context

This adds for support for configuring the integration to use an AWS managed metric stream within the same configured account.

## Changes

- Added new field into schema
- Added docs update


signalfx/resource_signalfx_aws_integration.go
```diff
@@ -248,6 +249,13 @@ func integrationAWSResource() *schema.Resource {
 				Default:     false,
 				Description: "Indicates that Splunk Observability should only sync recommended statistics",
 			},
+			"metric_streams_managed_externally": {
+				Type:     schema.TypeString,
```
bgola-signalfx: `schema.TypeBool` ?

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -302,41 +310,23 @@ func awsIntegrationAPIToTF(d *schema.ResourceData, aws *integration.AwsCloudWatc
...
-	if err := d.Set("enable_check_large_volume", aws.EnableCheckLargeVolume); err != nil {
-		return err
-	}
-	if err := d.Set("sync_custom_namespaces_only", aws.SyncCustomNamespacesOnly); err != nil {
-		return err
-	}
-	if err := d.Set("named_token", aws.NamedToken); err != nil {
-		return err
-	}
+	err := multierr.Combine(
```
bgola-signalfx, line 314: Nice!

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -248,6 +249,13 @@ func integrationAWSResource() *schema.Resource {
 				Default:     false,
 				Description: "Indicates that Splunk Observability should only sync recommended statistics",
 			},
+			"metric_streams_managed_externally": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  false,
+				Description: "If set to true, Splunk Observability Cloud accepts data from Metric Streams managed from the AWS console. " +
+					"The AWS account sending the Metric Streams and the AWS account in the Splunk Observability Cloud integration have to match.",
```
bgola-signalfx: Maybe it is worth to add here and in the markdown that this field is effective only when `use_metric_streams_sync = true`?

signalfx/resource_signalfx_aws_integration.go
MovieStoreGuy: Good catch, I completely missed that!

signalfx/resource_signalfx_aws_integration.go
MovieStoreGuy: It made sense, and it brought me joy.

bgola-signalfx (approved):

PR #579, opened 2025-03-14
Feature: Adding in support for encoding messages for AWS

MovieStoreGuy:
## Context

I am adding the encode functionality for the aws integration to get read for the resource implementation.
Since the integration is has significant number of fields, I have split this up for simpler review

## Changes

- Adding encoding for aws integration.

### DIFF
--- a/internal/convert/convert_any.go
+++ b/internal/convert/convert_any.go
@@ -1,0 +1,9 @@
+// Copyright Splunk, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package convert
+
+// ToAny provides a way to convert back to the generic terraform values
+func ToAny[T any](in T) any {
+	return in
+}
--- a/internal/definition/awsintegration/schema_integration.go
+++ b/internal/definition/awsintegration/schema_integration.go
@@ -321,8 +321,8 @@ func decodeTerraform(rd *schema.ResourceData) (*integration.AwsCloudWatchIntegra
 	return cwi, nil
 }
 
-func encodeTerraform(aws *integration.AwsCloudWatchIntegration, d *schema.ResourceData) (errs error) {
-	err := multierr.Combine(
+func encodeTerraform(aws *integration.AwsCloudWatchIntegration, d *schema.ResourceData) error {
+	errs := multierr.Combine(
 		d.Set("integration_id", aws.Id),
 		d.Set("name", aws.Name),
 		d.Set("enabled", aws.Enabled),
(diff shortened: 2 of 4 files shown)


internal/convert/convert_any.go
```diff
@@ -0,0 +1,9 @@
+// Copyright Splunk, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package convert
+
+// ToAny provides a way to convert back to the generic terraform values
+func ToAny[T any](in T) any {
+	return in
+}
```
bgola-signalfx, line 9: What's a benefit of this compared to `(any)(in)`?

internal/definition/awsintegration/schema_integration.go
```diff
@@ -319,3 +320,91 @@ func decodeTerraform(rd *schema.ResourceData) (*integration.AwsCloudWatchIntegra
 
 	return cwi, nil
 }
+
+func encodeTerraform(aws *integration.AwsCloudWatchIntegration, d *schema.ResourceData) (errs error) {
+	err := multierr.Combine(
```
bgola-signalfx: nit: can we assign to `errs` here and get rid of the `errs = multierr.Append(errs, err)` line?

internal/convert/convert_any.go
MovieStoreGuy: I honestly just got tired of typing it, but this helps going from the typed values into terraform expected types since you can't cast `[]T` into `[]any` which would be neat.

internal/definition/awsintegration/schema_integration.go
```diff
@@ -319,3 +320,90 @@ func decodeTerraform(rd *schema.ResourceData) (*integration.AwsCloudWatchIntegra
...
+		d.Set("use_metric_streams_sync", aws.MetricStreamsSyncState == "ENABLED"),
+		d.Set("enable_logs_sync", aws.LogsSyncState == "ENABLED"),
+		d.Set("enable_check_large_volume", aws.EnableCheckLargeVolume),
+		d.Set("sync_custom_namespaces_only", aws.SyncCustomNamespacesOnly),
+		d.Set("named_token", aws.NamedToken),
+		d.Set("collect_only_recommended_stats", aws.CollectOnlyRecommendedStats),
+		d.Set("metric_streams_managed_externally", aws.MetricStreamsManagedExternally),
+	)
+
+	if aws.Token != "" {
```
bgola-signalfx, line 341: Hmm, I thought I commented on this last week but it looks like I didn't...

Anyway, shouldn't we check for `aws.Key` too?

MovieStoreGuy: Fixed in https://github.com/splunk-terraform/terraform-provider-signalfx/pull/580

MovieStoreGuy: Added :D

bgola-signalfx (approved):

PR #595, opened 2025-04-02
Enhancement: Make api_key WriteOnly

mistastn:
## Context

Since Terraform 1.11 there are write only arguments that do not get stored in the state https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only

This is a perfect usecase for this to not leak the API keys in the state file. There are other places this could/should be used but I want to go small. This also means that Terraform 1.11 is required for this functionality.

## Changes

- Enable WriteOnly for PageDuty API key


mistastn:

MovieStoreGuy (approved):

PR #605, opened 2025-05-06
Feature: project tracking

MovieStoreGuy:
## Context

This allows users to understand where there resources are being created and link them to a git project automatically.

## Changes

- Added new preview flag `provider.track`, it is disabled by default and required to opt in
- Added ability to include project tags on resources being created
- Fixed a bug within detectors trying to use experimental feature.


signalfx/resource_signalfx_detector.go
```diff
@@ -538,7 +538,7 @@ func detectorCreate(d *schema.ResourceData, meta any) error {
 	payload.Teams = pmeta.MergeProviderTeams(
 		context.TODO(),
 		meta,
-		payload.Tags,
```
MovieStoreGuy, line 541: Noticed the issue as part of testing it locally, fixing in the same PR to avoid issues later.

atoulme (approved):

PR #606, opened 2025-05-06
Update data_source.go | Use id (member ID) instead of userId as per SignalFx API documentation

anuragmuley:
Updated the code to use id (member ID) instead of userId in accordance with the SignalFx Observability API documentation. The id field uniquely identifies a member and should be used in all references. This ensures consistency and aligns with the latest API expectations. Reference: 
https://dev.splunk.com/observability/reference/api/organizations/latest


anuragmuley:

MovieStoreGuy (approved):

PR #614, opened 2025-06-11
AGG-2000 Add terraform calling logic for Automated Archival APIs

ddanuj:
Pull request for the golang impl of Automated Archival APIs - https://github.com/signalfx/signalfx-go/pull/247


ddanuj:

ddanuj:

ddanuj:

go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/splunk-terraform/terraform-provider-signalfx
 
-go 1.24.1
+go 1.23.0
```
MovieStoreGuy: Why is go being downgraded here?

internal/definition/autoarchiveexemptmetric/resource.go
```diff
@@ -0,0 +1,95 @@
...
+
+	exempt_metrics, err := sfx.CreateExemptMetrics(ctx, details)
+	if err != nil {
+		return tfext.AsErrorDiagnostics(common.HandleError(ctx, err, data))
+	}
+
+	// Set the ID of the resource to the string representation of the ids of all the exempt metrics
+	if len(*exempt_metrics) == 0 {
+		return tfext.AsErrorDiagnostics(common.HandleError(ctx, errors.New("no exempt metrics found after creation"), data))
+	}
```
MovieStoreGuy, line 68: Do we know before hand if no exempt metrics will be created?

internal/definition/autoarchiveexemptmetric/resource.go
```diff
@@ -0,0 +1,95 @@
...
+	// Join the IDs of the exempt metrics into a single string, separated by commas
+	ids := make([]string, len(*exempt_metrics))
+	for i, metric := range *exempt_metrics {
+		ids[i] = *metric.Id
+	}
+	data.SetId(strings.Join(ids, ","))
```
MovieStoreGuy, line 74: Do these values need to be sorted to ensure consistency? To avoid any potential misleading terraform diffs?

internal/definition/autoarchiveexemptmetric/resource.go
```diff
@@ -0,0 +1,95 @@
...
+	return tfext.AsErrorDiagnostics(encodeTerraform(exempt_metrics, data))
+}
+
+func resourceDelete(ctx context.Context, data *schema.ResourceData, meta any) diag.Diagnostics {
+	sfx, err := pmeta.LoadClient(ctx, meta)
+	if err != nil {
+		return tfext.AsErrorDiagnostics(err)
+	}
+
+	if len(data.State().ID) == 0 {
```
MovieStoreGuy: It is technically faster here to check if it is equal to `""` instead of length.

internal/definition/autoarchiveexemptmetric/resource.go
MovieStoreGuy: Actually, considering that this can only be applied per org id, any reason not to set it to the org ID?

ddanuj:

internal/definition/autoarchivesettings/schema.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+		"enabled": {
+			Type:        schema.TypeBool,
+			Required:    true,
+			Description: "Whether the automated archival is enabled for this organization or not",
+		},
+		"lookback_period": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "This tracks if a metric was unused in the past N number of days (N one of 30, 45, or 60). We’ll archive a metric if it wasn’t used in the lookback period. The value here uses ISO 8061 duration format. Examples - 'P30D', 'P45D', 'P60D'",
```
MovieStoreGuy, line 51: Why are using an alternative relative time schema here compared to what has already been established other components?

internal/definition/autoarchivesettings/schema.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+		"grace_period": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Grace period is an org level setting that applies to the newly created metrics. This allows customers to protect newly added metrics that users haven’t had the time to use in charts and detectors from being automatically archived The value here uses ISO 8061 duration format. Examples - 'P0D', 'P15D', 'P30D', 'P45D', 'P60D'",
```
MovieStoreGuy, line 56: Same again here with regards to the time formatting used here.

go.mod
ddanuj: I think I didn't have this go version installed. But I don't remember setting this explicitly - Anyways fixed in the next commit

internal/definition/autoarchiveexemptmetric/resource.go
ddanuj: ideally this should not happen. Either the exempt metrics will be created, or there will be an error thrown which is handled before this line. But to protect the code below that joins all the exempt metric ids together, I added this code

internal/definition/autoarchiveexemptmetric/resource.go
ddanuj: I see that we use this id to delete the same resource. Delete API for exempt metrics takes a list of metric ids (and not the organization id). So I wanted to store the list of metric ids created. Since the resource id can only be a string, I thought a string representation of the list of ids would work here.

internal/definition/autoarchiveexemptmetric/resource.go
ddanuj: Fixed in the next commit

internal/definition/autoarchivesettings/schema.go
ddanuj: can you please elaborate on "alternative relative schema"? What has been established by other components? Example?

PR #617, opened 2025-06-16
Add an optional flag to skip saving the org token to state

ar-cisco:
## Summary

This PR introduces a feature for SignalFx organization tokens by adding:

- Optional secret storage control - A new store_secret parameter that allows users to prevent token secrets from being stored in Terraform state
- Token lookup data source - A new signalfx_org_token_lookup data source for retrieving existing token secrets

### Functionality Enhancement - `store_secret` Parameter:
- Added a new optional `store_secret` boolean parameter to the signalfx_org_token resource
- Defaults to `true` for backward compatibility
- When set to`false`, the token secret is not stored in Terraform state, allowing the resource to be used for org token settings alone (e.g. per token limits)
- Modified `orgTokenAPIToTF` function to conditionally store the secret based on this setting

### New Data Source - `signalfx_org_token_lookup`
- Added signalfx_org_token_lookup data source to allow lookup of existing token secrets
- Provides a way to retrieve token secrets when store_secret is set to false on the resource
- Registered the new data source in the provider configuration

### Test Coverage

- `TestAccOrgTokenSecureStorage` - Tests tokens with store_secret = false
- `TestAccOrgTokenBackwardCompatibility` - Ensures existing behaviour is preserved
- `TestAccOrgTokenSecureStorageWithDPMLimits` - Tests secure storage with DPM limits
- `TestAccOrgTokenSecureStorageWithLimits` - Tests secure storage with usage limits
- `TestAccOrgTokenExplicitSecretStorage` - Tests explicit store_secret = true
- `TestAccOrgTokenStoreSecretToggle` - Tests toggling the store_secret parameter

### Inconsistency Check
- Replaced `fmt` with `log` for one of the debug logs to be consistent with the other logging.

## Benefits
- Enhanced Security - Users can now prevent token secrets from being stored in Terraform state files
- Backward Compatibility - Existing configurations continue to work without changes

## Examples
```terraform
# Secure token without secret storage
resource "signalfx_org_token" "secure_token" {
  name         = "MySecureToken"
  description  = "Token with no secret storage"
  store_secret = false
  auth_scopes  = ["API"]
}

# Lookup token secret when needed
data "signalfx_org_token_lookup" "my_token" {
  name = "MySecureToken"
}
```

### DIFF
--- a/signalfx/resource_signalfx_org_token.go
+++ b/signalfx/resource_signalfx_org_token.go
@@ -151,7 +151,7 @@ func orgTokenResource() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     true,
-				Description: "Whether to store the token's secret in the terraform state. Set to false for improved security. Defaults to true for backward compatibility.",
+				Description: "Whether to store the token's secret in the terraform state. Defaults to true.",
 			},
 		},
 
--- a/signalfx/resource_signalfx_org_token.go
+++ b/signalfx/resource_signalfx_org_token.go
@@ -147,6 +147,12 @@ func orgTokenResource() *schema.Resource {
 				Computed:  true,
 				Sensitive: true,
 			},
+			"store_secret": &schema.Schema{
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     true,
+				Description: "Whether to store the token's secret in the terraform state. Set to false for improved security. Defaults to true for backward compatibility.",
+			},
 		},
 
 		Create: orgTokenCreate,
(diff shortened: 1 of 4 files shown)


ar-cisco:
recheck

MovieStoreGuy:
Hey @ar-cisco ,

Unfortunately you need to copy and paste the text in full for it to count as you acknowledging the CLA.

signalfx/resource_signalfx_org_token.go
```diff
@@ -347,6 +353,49 @@ func orgTokenAPIToTF(d *schema.ResourceData, t *orgtoken.Token) error {
 		return err
 	}
 
+	secret := func() string {
+		if d.Get("store_secret").(bool) {
+			return t.Secret
+		}
+		return ""
+	}()
```
MovieStoreGuy, line 361: I would prefer not needing to lambda function here in order to get the desired behaviour since it makes the stack frames harder to read if sent to support.

signalfx/resource_signalfx_org_token.go
```diff
@@ -347,6 +353,49 @@ func orgTokenAPIToTF(d *schema.ResourceData, t *orgtoken.Token) error {
...
+
+	if err := d.Set("secret", secret); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func orgTokenLookupDataSource() *schema.Resource {
+	return &schema.Resource{
```
MovieStoreGuy, line 371: Can you provide a description for the resource itself?

MovieStoreGuy (commented):
The code is fine, it doesn't require any 1.0+ terraform features which is great.

I only ask that you review the failing checks and address any of the concerns there.

MovieStoreGuy:
I am going to reject this change since I'd like to adopt private read only vars that only accessible in v1.11.0+, and also re-evaluate how this component manages secrets itself.

PR #623, opened 2025-07-03
 AGG-2000 Provider logic for managing Automated Archival resources

ddanuj:
Pull request for the golang impl of Automated Archival APIs - https://github.com/signalfx/signalfx-go/pull/247


internal/definition/autoarchiveexemptmetric/schema.go
```diff
@@ -0,0 +1,132 @@
...
+			Computed:    true,
+			Description: "Timestamp of when the automated archival setting was last updated",
+		},
+		"name": {
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Name of the metric to be exempted from automated archival",
+		},
+	}
+)
```
MovieStoreGuy: I would like to avoid having globally defined variables were possible. 

The reason being is that it increases the idle memory needed for the provided which may already being more than it needs due the terraform frameworks and such. This allows for users to continue to use low memory CI/CD pipelines where applicable.

internal/definition/autoarchiveexemptmetric/schema.go
```diff
@@ -0,0 +1,132 @@
...
+
+				*exempt_metrics = append(*exempt_metrics, exempt_metric)
+			}
+		}
+	}
+	return exempt_metrics, nil
+}
+
+func encodeTerraform(exempt_metrics *[]automated_archival.ExemptMetric, data *schema.ResourceData) error {
+	if exempt_metrics == nil {
```
MovieStoreGuy: Would this be the same as checking of the length of the array is nil?

internal/definition/autoarchiveexemptmetric/schema.go
```diff
@@ -0,0 +1,132 @@
...
+		if metric.LastUpdated != nil {
+			exemptMetricsList[i]["last_updated"] = *metric.LastUpdated
+		}
+	}
+
+	errs := multierr.Combine(
+		data.Set("exempt_metrics", exemptMetricsList),
+	)
+
+	return errs
```
MovieStoreGuy: You can simplify this to be:

```golang
return data.Set("exempt_metrics", exemptMetricsList)
```

internal/definition/autoarchiveexemptmetric/schema.go
```diff
@@ -0,0 +1,132 @@
...
+			Type:        schema.TypeString,
+			Required:    true,
+			Description: "Name of the metric to be exempted from automated archival",
+		},
+	}
+)
+
+func newSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"exempt_metrics": {
```
MovieStoreGuy: Should this set a min length of 1?

internal/definition/autoarchiveexemptmetric/schema.go
```diff
@@ -0,0 +1,132 @@
...
+				Schema: exemptMetricSchema,
+			},
+			ForceNew: true,
+		},
+	}
+}
+
+func decodeTerraform(data *schema.ResourceData) (*[]automated_archival.ExemptMetric, error) {
+	exempt_metrics := &[]automated_archival.ExemptMetric{}
+	if v, ok := data.GetOk("exempt_metrics"); ok {
```
MovieStoreGuy: Can I suggest inverting the condition here? The amount of nesting can be reduced significantly.

internal/definition/autoarchiveexemptmetric/schema.go
```diff
@@ -0,0 +1,132 @@
...
+			if itemMap, ok := item.(map[string]any); ok {
+				exempt_metric := automated_archival.ExemptMetric{
+					Name: itemMap["name"].(string),
+				}
+
+				// Only set if field exists and contains a value
+				if creator, exists := itemMap["creator"]; exists && creator != nil && creator.(string) != "" {
+					str := creator.(string)
+					exempt_metric.Creator = &str
+				}
```
MovieStoreGuy: This can be simplified by trying to cast the type instead of checking if exists.

```golang
if thing, ok := m["key"].(string); ok && thing != "" {
  // Use thing
}
```
Reduces the amount of checks needed and simplifies the code a bit more.

MovieStoreGuy (commented):
Just a few minor nits from me, however, I do have one outstanding question with regards to mixed usage of both UI and terraform. What is the anticipated usage pattern or adoption users should take?

internal/definition/autoarchiveexemptmetric/schema.go
ddanuj: replacing the variable with a function

internal/definition/autoarchiveexemptmetric/schema.go
ddanuj: the check here checks if the variable supplied as the function parameter has not been initialized or explicitly set to nil. But I also want to check if the length is 0. So I'll add that as well

internal/definition/autoarchiveexemptmetric/schema.go
ddanuj: Fixed

internal/definition/autoarchiveexemptmetric/schema.go
ddanuj: Fixed

internal/definition/autoarchiveexemptmetric/schema.go
ddanuj: Fixed

internal/definition/autoarchiveexemptmetric/schema.go
ddanuj: Fixed

ddanuj:
> Just a few minor nits from me, however, I do have one outstanding question with regards to mixed usage of both UI and terraform. What is the anticipated usage pattern or adoption users should take?

I discussed this with the team about the use of this provider when there are UI components already for doing the same thing.

Some folks mentioned that few customers need Infrastructure as a Code structure to manage their resources. So they prefer terraform to the UI to manage these admin configs (both AA settings and exempt metrics are to be used by admins)

Some also mentioned that even us (Splunk devs) should be using Terraform more to manage our own resources - not just AA settings and exempt metrics but also metric rulesets, detectors etc. Terraform has state which is better than UI which can be manipulated by multiple users using common credentials.

go.mod
```diff
@@ -2,10 +2,12 @@ module github.com/splunk-terraform/terraform-provider-signalfx
 
 go 1.24.1
 
+toolchain go1.24.4
+
 require (
 	github.com/bgentry/go-netrc v0.0.0-20140422174119-9fd32a8b3d3d
 	github.com/davecgh/go-spew v1.1.1
-	github.com/go-git/go-git/v5 v5.16.2
+	github.com/go-git/go-git/v5 v5.14.0
```
MovieStoreGuy: This should not be downgraded, any reason why it has been?

MovieStoreGuy (approved):

go.mod
ddanuj: not sure - let me restore

ddanuj: Fixed

Issue #464, opened 2024-02-20
URL is not correct when running terraform state show resource_type.resource_id


bran1501:
Hi Team, 

I've been adding some resources to my environment and I've configured the provider info correctly, check this out
```
terraform {
  required_providers {
    signalfx = {
      source = "splunk-terraform/signalfx"
      version = "9.0.1"
    }
  }
}
provider "signalfx" {
  auth_token = var.access_token
  api_url    = "https://api.${var.realm}.signalfx.com"
}

variable "access_token" {
  description = "Splunk Access Token"
  default = "xxxxxx"
}

variable "realm" {
  description = "Splunk Realm"
  default     = "us1"
}
```

This is my resource 
```
resource "signalfx_heatmap_chart" "tfchart_08" {
    name = "CB_TF_p99Latency"
    description = "99th percentile of latency by each service"
    program_text = <<-EOF
        A = data("dem_latency").percentile(99, by=['dem_service']).publish(label="latency")
    EOF
  color_scale {
    gte   = 100
    color = "red"
  }
  color_scale {
    lt   = 30
    color = "green"
  }
  color_scale {
    lt    = 100
    gte   = 30
    color = "yellow"
  }
}
```
When I run terraform state show signalfx_heatmap_chart.tfchart_08
```
bran1501@localhost terraform_labs % terraform state show signalfx_heatmap_chart.tfchart_08
\# signalfx_heatmap_chart.tfchart_08:
resource "signalfx_heatmap_chart" "tfchart_08" {
    description      = "99th percentile of latency by each service"
    disable_sampling = false
    group_by         = []
    hide_timestamp   = false
    id               = "GGypvwdA4AA"
    name             = "CB_TF_p99Latency"
    program_text     = <<-EOT
        A = data("dem_latency").percentile(99, by=['dem_service']).publish(label="latency")
    EOT
    timezone         = "UTC"
    unit_prefix      = "Metric"
    url              = "https://app.signalfx.com/#/chart/GGypvwdA4AA"
    color_scale {
        color = "green"
        gt    = 340282346638528860000000000000000000000
        gte   = 340282346638528860000000000000000000000
        lt    = 30
        lte   = 340282346638528860000000000000000000000
    }
    color_scale {
        color = "red"
        gt    = 340282346638528860000000000000000000000
        gte   = 100
        lt    = 340282346638528860000000000000000000000
        lte   = 340282346638528860000000000000000000000
    }
    color_scale {
[...]

PR #626, opened 2025-07-18
Feature: Enable auto detect custom domains

MovieStoreGuy:
# Context

This will allow for the provider to infer the custom domain that is used for our customers instead of requiring a user to set it.

## Changes

- Deprecates `custom_app_url` within the provider
- Adds support for detecting the custom domain for the organization settings.


MovieStoreGuy:
Fixes https://github.com/splunk-terraform/terraform-provider-signalfx/issues/464

durgeshsplunk (approved):

PR #628, opened 2025-07-18
fix: pass alias into variable overrides in dashboard group

tnoff:
We're hitting the following error on the latest terraform version updating dashboards

```
{
  "code" : 400,
  "message" : "Empty alias is not allowed for variable filters.,Empty alias is not allowed for variable filters.,Empty alias is not allowed for variable filters.,Empty alias is not allowed for variable filters.,Empty alias is not allowed for variable filters.,Empty alias is not allowed for variable filters."
}
```

I've got a script to fix where I update the payload being sent over by terraform, it seems like you can pass just about anything into `alias` and it doesn't get used


tnoff:

MovieStoreGuy:
Hey @tnoff , 

I haven't had a proper look as to why this is happening, but I will check the code paths to see what is an acceptable fix here.

tnoff:
I believe this has since been fixed in the API, thanks all!

PR #634, opened 2025-07-30
Color Palette updated to reflect correct UI colors

durgeshsplunk:
### Context ### 

Currently only 16 colors for vizualization colors and 22 are accepted for color_scale within heatmap and histogram
This change will reflect correct colors on the UI. The change will make sure that colors that were previously used are not breaking in nature. This will however lead to not having the exact color for the ones that are no longer supported

A new color palette was added to support additional colors for color_scale and histogram color_theme.

Docs have been updated to reflect the change along with necessary tests.


durgeshsplunk:

MovieStoreGuy (approved):

PR #635, opened 2025-08-01
Feature: Unified chart component 

MovieStoreGuy:
# Context

In order to simplify charts within the provider with both usage, and development. They are being merged under one resource type. This allows us the minimise issues with sprawling options, keep the component clean, and also allow for a smoother migration since both the previous charts and the new one can exist together.

## Breaking Changes
- [ ] Requires adoption of terraform 1.2+ features which will mean users will need time to upgrade before rolling out

## Changes
- Adding new chart resource
- Adding new validators 

# Next Steps

Considering that testing isn't being performed on the PR yet, I will merge this change in stages but I wanted to gauge interested and feedback early.


MovieStoreGuy:
Closing this in favour of the other approach documented.

PR #639, opened 2025-08-11
Chore: Adding package for framework validation

MovieStoreGuy:
# Context

This package is intended to simplify the work being done within the components

## Changes

- Adding new package `fwtest` for component validation

### DIFF
--- a/internal/framework/fwtest/schema_validation.go
+++ b/internal/framework/fwtest/schema_validation.go
@@ -1,0 +1,80 @@
+// Copyright Splunk, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package fwtest
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"reflect"
+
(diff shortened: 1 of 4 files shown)


internal/framework/fwtest/schema_validation.go
```diff
@@ -0,0 +1,80 @@
...
+			continue
+		}
+		// TODO(MovieStoreGuy): This is fine until we handle more complex types.
+		switch f := reflect.Zero(field.Type).Interface().(type) {
+		case attr.Value:
+			expected[named] = f
+		default:
+			// Handle the case where the field is potentially nested or complex.
+			panic("You have entered the twilight zone. Please check the model definition.")
+		}
```
MovieStoreGuy, line 58: I purposely ignored this since I have no reason to make it more complicated at the moment, but I imagine that will come once nested models are adopted.

durgeshsplunk (approved):

PR #691, opened 2025-12-04
add cold_poll_rate option for AWS integration

kroman379:


signalfx/resource_signalfx_aws_integration.go
```diff
@@ -181,6 +181,11 @@ func integrationAWSResource() *schema.Resource {
 				Description:  "AWS poll rate (in seconds). Between `60` and `600`.",
 				ValidateFunc: validation.IntBetween(60, 600),
 			},
+			"cold_poll_rate": {
+				Type:        schema.TypeInt,
+				Optional:    true,
+				Description: "AWS cold poll rate (in seconds).",
+			},
```
bgola-signalfx, line 189:
```suggestion
				ValidateFunc: validation.IntBetween(60, 3600),
			},
```

signalfx/resource_signalfx_aws_integration.go
```diff
@@ -421,6 +426,12 @@ func awsIntegrationAPIToTF(d *schema.ResourceData, aws *integration.AwsCloudWatc
 		}
 	}
 
+	if aws.ColdPollRate != 0 {
+		if err := d.Set("cold_poll_rate", aws.ColdPollRate/1000); err != nil {
+			return err
+		}
+	}
+
```
bgola-signalfx: nit: I would move it to line 324 (closer to the regular poll rate)

bgola-signalfx (approved):

signalfx/resource_signalfx_aws_integration.go
kroman379: Do we want to use the current max cold poll rate (20m) or the target one (60m)?

bgola-signalfx: I think we can use the target one here as API is going to double check this anyway

bgola-signalfx (approved):

anosek-an (approved):

PR #692, opened 2025-12-04
DASHM-399: Dashify Template Resource to Beta Release

jafaas:
New terraform resource to enable syncing of SignalFX Dashify Templates.
This includes Create, Update, Delete, and Import.

cc: @MovieStoreGuy 


jafaas:

Recheck

jafaas:

jafaas:
Recheck

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+
+type resourceDashifyTemplateModel struct {
+	Id               types.String `tfsdk:"id"`
+	TemplateContents types.String `tfsdk:"template_contents"`
+}
+
+var (
+	_ resource.Resource                = &ResourceDashifyTemplate{}
+	_ resource.ResourceWithConfigure   = &ResourceDashifyTemplate{}
+	_ resource.ResourceWithImportState = &ResourceDashifyTemplate{}
```
MovieStoreGuy:
```go
	_ resource.Resource                = (*ResourceDashifyTemplate)(nil)
	_ resource.ResourceWithConfigure    = (*ResourceDashifyTemplate)(nil)
    _ resource.ResourceWithImportState = (*ResourceDashifyTemplate)(nil)
```

They should use nil instead of defining values onto into the global stack.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+
+func (r *ResourceDashifyTemplate) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Manage Dashify templates for modern dashboards in Splunk Observability Cloud",
+		Attributes: map[string]schema.Attribute{
+			"id": fwshared.ResourceIDAttribute(),
+			"template_contents": schema.StringAttribute{
+				Required:    true,
+				Description: "JSON contents of the dashify template. Must be valid JSON containing the template structure with metadata and spec sections.",
+			},
```
MovieStoreGuy, line 61: Is there a schema for the template contents that could be expanded here to make it easier to validate locally instead of waiting for the apply operations?

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+	templateContents := model.TemplateContents.ValueString()
+	var js interface{}
+	err := json.Unmarshal([]byte(templateContents), &js)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Invalid JSON",
+			fmt.Sprintf("template_contents must be valid JSON: %s", err.Error()),
+		)
+		return
+	}
```
MovieStoreGuy, line 83: This only validates that it is valid JSON, is there anyway that the schema could be validated here as well?

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Make API request
+	details := r.Details()
+	url := fmt.Sprintf("%s%s", details.APIURL, TemplateAPIPath)
```
MovieStoreGuy: Please don't use `sprintf` for concatenating paths together, use `path.Join` or `filepath.Join`

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+
+	httpReq, err := http.NewRequestWithContext(ctx, "POST", url, bytes.NewBuffer([]byte(templateContents)))
```
MovieStoreGuy: Use `http.MethodPost` instead of the raw string `POST`.

MovieStoreGuy: I would also suggest using `strings.NewReader` instead of creating a bytes buffer.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+			fmt.Sprintf("Could not create HTTP request: %s", err.Error()),
+		)
+		return
+	}
+
+	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("X-SF-TOKEN", details.AuthToken)
+
+	client := &http.Client{}
+	httpResp, err := client.Do(httpReq)
```
MovieStoreGuy, line 102: You should be using the sfx go sdk instead of the constructing the raw payload here.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+	defer httpResp.Body.Close()
+
+	body, err := io.ReadAll(httpResp.Body)
+	if err != nil {
+		resp.Diagnostics.AddError(
+			"Error Reading Response",
+			fmt.Sprintf("Could not read response body: %s", err.Error()),
+		)
+		return
+	}
```
MovieStoreGuy, line 119: I would do two things here:
1. Check the status code first since you can short cut some things here.
1. Use a JSON decoder so you can stream the contents internally instead of reading everything into memory to then pass back to the json marshaler. 

To check the status code you can do something like:
```go
switch resp.StatusCode {
case http.StatusOK, http.StatusCreated:
  // Expected case, nothing to do here
default:
  // Handle error here
}
```

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+		resp.Diagnostics.AddError(
+			"Error Creating Template",
+			fmt.Sprintf("API returned status %d: %s", httpResp.StatusCode, string(body)),
+		)
+		return
+	}
+
+	// Parse response to get the template ID
+	// The API returns: {"data": {"id": "...", ...}, "errors": [], "includes": []}
+	var result map[string]interface{}
```
MovieStoreGuy, line 131: Use `any` over `interface{}`

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+	if err := json.Unmarshal(body, &result); err != nil {
+		resp.Diagnostics.AddError(
+			"Error Parsing Response",
+			fmt.Sprintf("Could not parse response: %s", err.Error()),
+		)
+		return
+	}
+
+	// Extract the data object
+	data, ok := result["data"].(map[string]interface{})
```
MovieStoreGuy, line 141: I would suggested using a typed value instead of a map so you can get the field validation done as part of the unmarshal step.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Error Parsing Response",
+			fmt.Sprintf("template ID not found in response data: %s", string(body)),
+		)
+		return
+	}
+
+	model.Id = types.StringValue(templateID)
+	tflog.Debug(ctx, "Created Dashify Template", map[string]interface{}{"id": templateID})
```
MovieStoreGuy, line 161: The extensions package `tfext` will generate the map for you with type support.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+		return
+	}
+
+	resp.Diagnostics.Append(resp.State.Set(ctx, &model)...)
+}
+
+func (r *ResourceDashifyTemplate) readTemplate(ctx context.Context, model *resourceDashifyTemplateModel, diags *diag.Diagnostics) {
+	details := r.Details()
+	templateID := model.Id.ValueString()
+	url := fmt.Sprintf("%s%s/%s", details.APIURL, TemplateAPIPath, templateID)
```
MovieStoreGuy, line 192: Please use `path.Join` or `filepath.Join` instead of `Sprintf`

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+
+	httpReq, err := http.NewRequestWithContext(ctx, "GET", url, nil)
```
MovieStoreGuy:
```suggestion
httpReq, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
```

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+			"Error Creating HTTP Request",
+			fmt.Sprintf("Could not create HTTP request: %s", err.Error()),
+		)
+		return
+	}
+
+	httpReq.Header.Set("X-SF-TOKEN", details.AuthToken)
+
+	client := &http.Client{}
+	httpResp, err := client.Do(httpReq)
```
MovieStoreGuy, line 206: This should be moved to the go sdk, not done in place.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+		return
+	}
+
+	if httpResp.StatusCode != http.StatusOK {
+		diags.AddError(
+			"Error Reading Template",
+			fmt.Sprintf("API returned status %d: %s", httpResp.StatusCode, string(body)),
+		)
+		return
+	}
```
MovieStoreGuy, line 237: Make this check in the same step as line 215 to handle all error handling early

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,398 @@
...
+	// Only keep the fields that users provide: metadata, spec, title, type (if user-provided)
+	filteredData := make(map[string]interface{})
+
+	// These are the fields users provide in their template
+	userProvidedFields := []string{"metadata", "spec", "title", "type"}
+	for _, field := range userProvidedFields {
+		if val, exists := templateData[field]; exists {
+			filteredData[field] = val
+		}
+	}
```
MovieStoreGuy, line 269: This should be moved to a concrete type instead of checking if a field exists.

MovieStoreGuy (commented):
Adding my comments for now, this should be a lot less noisy once the PR has gone in to sync main back to the release branch.

My review is somewhat incomplete since there is a lot of points that are repeated across the different methods.

internal/framework/dashify/resource_dashify_template.go
jafaas: The schema validation is not clear at this time.

internal/framework/dashify/resource_dashify_template.go
jafaas: The schema validation is not clear at this time.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,399 @@
...
+	r.readTemplate(ctx, &model, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	if model.Id.IsNull() {
+		// Template was deleted
+		resp.State.RemoveResource(ctx)
+		return
+	}
```
MovieStoreGuy, line 184: I believe that this isn't required with the current import by ID, but I need to confirmed.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,399 @@
...
+		)
+		return
+	}
+	defer httpResp.Body.Close()
+
+	if httpResp.StatusCode == http.StatusNotFound {
+		tflog.Debug(ctx, "Dashify Template not found", map[string]interface{}{"id": templateID})
+		model.Id = types.StringNull()
+		return
+	}
```
MovieStoreGuy, line 220: I have a feeling this may not be needed with the import via ID, but I need to double check again.

internal/framework/dashify/resource_dashify_template.go
```diff
@@ -0,0 +1,399 @@
...
+
+	body, err := io.ReadAll(httpResp.Body)
+	if err != nil {
+		diags.AddError(
+			"Error Reading Response",
+			fmt.Sprintf("Could not read response body: %s", err.Error()),
+		)
+		return
+	}
```
MovieStoreGuy, line 229: Push this to use  `json.NewDecoder` instead to reduce overall required buffer.

MovieStoreGuy (commented):
Before moving to GA, look to move some of the repeated functionality to an SDK, validate the read, and general go practices cleanup.

PR #713, opened 2026-03-12
Feature: Adding Datasource for built in content

MovieStoreGuy:
# Context

This change will allow for users to reference the built in content available to them in their own dashboard groups.

## Changes
- Added `signalfx_builtin_dashboards` datasource


nvarun (approved):

anosek-an (approved):
This is great work, looking forward to using that feature!

PR #717, opened 2026-03-17
Remove mentions of AWS log collection

akonarska:


internal/definition/awsintegration/schema_integration.go
```diff
@@ -177,12 +177,6 @@ func newIntegrationSchema() map[string]*schema.Schema {
 			Computed:    true,
 			Description: "Enables the use of Cloudwatch Metric Streams for metrics synchronization.",
 		},
-		"enable_logs_sync": {
-			Type:        schema.TypeBool,
-			Optional:    true,
-			Computed:    true,
-			Description: "Enables AWS logs synchronization.",
-		},
```
MovieStoreGuy, line 185: Can you mark the field as deprecated instead of removing it?

We can remove it in v10 once that is finished.

MovieStoreGuy (commented):
Just one change from me :D

MovieStoreGuy (approved):

PR #730, opened 2026-04-02
Feature: Added support for Customized Auto Detectors

MovieStoreGuy:
# Context

This adds support for customized auto detector but depends on an upgraded sdk so is targeted to land in v10 officially.

# Changes
- Updated the autodetect datasource to return fields that are configurable in the resource
- Added resource as part of the changes.


signalfx/resource_signalfx_aws_integration.go
```diff
@@ -197,13 +197,6 @@ func integrationAWSResource() *schema.Resource {
 				Computed:    true,
 				Description: "Enables the use of Cloudwatch Metric Streams for metrics synchronization.",
 			},
-			"enable_logs_sync": {
-				Type:        schema.TypeBool,
-				Optional:    true,
-				Computed:    true,
-				Description: "Enables AWS logs synchronization.",
-				Deprecated:  "enable_logs_sync field is no longer used and will be removed in a next major release.",
```
atoulme, line 205: likely a separate change

atoulme (approved):

signalfx/resource_signalfx_aws_integration.go
MovieStoreGuy: Yes, I have just messed up some git history and will need to sync it back to the release branch.

PR #736, opened 2026-05-07
Chore: Upgrading terraform SDK changes

MovieStoreGuy:


GNUmakefile
```diff
@@ -52,7 +52,7 @@ checklicense: $(ADDLICENCESE)
 
 .PHONY: govulncheck
 govulncheck: $(GOVULNCHECK)
-	$(GOVULNCHECK)
+	$(GOVULNCHECK) ./...
```
atoulme, line 55: Is this change intended?

atoulme (approved):

GNUmakefile
MovieStoreGuy: Yeah, it was required with the latest version of vulncheck

PR #742, opened 2026-06-04
Add Email notification cc/bcc support and bump signalfx-go to v1.59.0.

manishaga:
Extend detector and SLO notification strings with optional cc/bcc lists, wire through API conversion, and document the new formats.


manishaga:

internal/common/notification_string.go
```diff
@@ -159,3 +160,69 @@ func NewNotificationFromString(str string) (*notification.Notification, error) {
...
+		return nil, nil
+	}
+	return addrs, nil
+}
+
+func formatEmailRecipientList(addrs []string) string {
+	if len(addrs) == 0 {
+		return ""
+	}
+	sorted := append([]string(nil), addrs...)
```
MovieStoreGuy: Use `slices.Clone` instead here.

internal/common/notification_string.go
```diff
@@ -159,3 +160,69 @@ func NewNotificationFromString(str string) (*notification.Notification, error) {
...
+			continue
+		}
+		if _, err := mail.ParseAddress(part); err != nil {
+			return nil, err
+		}
+		addrs = append(addrs, part)
+	}
+	if len(addrs) == 0 {
+		return nil, nil
+	}
```
MovieStoreGuy, line 216: If you have reached this point with no valid entries, like, `cc||||` is that a valid usecase?

MovieStoreGuy (approved):

internal/common/notification_string.go
manishaga: done

MovieStoreGuy: It looks like this was missed,  do you mind providing an update.

manishaga: fixed

PR #754, opened 2026-07-07
Fix ColorThemeIndex validation for time chart

JoelleJS:
The wrong color palette is used to validate the HistogramChartOptions returned by the API, resulting in an error when the API returns a ColorThemeIndex higher than 15: `invalid color theme index: 16`


JoelleJS:

durgeshsplunk (approved):

PR #797, opened 2026-07-24
ALEN-2537 Adds Terraform provider support for managing Splunk Observability Cloud email templates through a new signalfx_email_template resource.

manishaga:
Summary
Adds Terraform provider support for managing Splunk Observability Cloud email templates through a new signalfx_email_template resource.
Changes
Added signalfx_email_template framework resource with CRUD support.
Registered the resource in the framework provider.
Bumped github.com/signalfx/signalfx-go to v1.63.0 for email template API support.
Added unit coverage for metadata, schema, create, read, update, and delete paths.
Added Terraform example, docs template, and generated docs page.


internal/framework/alert/testdata/01_email_template_updated.tf
```diff
@@ -0,0 +1,18 @@
...
+
+  to = [
+    "primary@example.com",
+    "secondary@example.com",
+  ]
+  cc = ["team@example.com"]
+
+  custom_headers = {
+    X-SFX-Template = "detector-updated"
+  }
```
MovieStoreGuy, line 14: What do custom headers do here?

internal/framework/alert/resource_email_template.go
```diff
@@ -0,0 +1,284 @@
...
+	return types.MapValueFrom(ctx, types.StringType, values)
+}
+
+func optionalEmailTemplateStringValue(value string) types.String {
+	if value == "" {
+		return types.StringNull()
+	}
+
+	return types.StringValue(value)
+}
```
MovieStoreGuy: These looks like these could reusable by other types, I would suggest moving into a shared library.

internal/framework/alert/resource_email_template_test.go
```diff
@@ -0,0 +1,169 @@
...
+		CustomHeaders: types.MapNull(types.StringType),
+	}))
+}
+
+func TestResourceEmailTemplateUnitTest(t *testing.T) {
+	t.Parallel()
+
+	var (
+		current emailtemplate.EmailTemplate
+		mu      sync.Mutex
```
MovieStoreGuy: There should not be a need for a mutex within a test

MovieStoreGuy: Each test should be atomic and used shared state.

go.sum
MovieStoreGuy, line 1: You'll need to do a go mod clean to ensure the older version is not picked up as part of the items to download.

internal/framework/alert/resource_email_template_test.go
manishaga: done

internal/framework/alert/resource_email_template.go
manishaga: done

internal/framework/alert/testdata/01_email_template_updated.tf
manishaga: done

MovieStoreGuy (approved):

